PR #29, opened 2022-02-10
fix git clone instructions

cmungall:


joeflack4:
@cmungall Chris! Hi there. Thanks for the fix!

PR #34, opened 2022-02-23
VSAC Wrangler updates: Appending last digits of OID to name

joeflack4:
## Description
In VSAC Wrangler, when choosing `palantir-concept-set-tables` for `--output-structure`, it will now mutate the value set names in cases where there are multiple value sets with the same name. In that case, it will append the last digits of the OID to the value set name.

## Update message
```
VSAC Wrangler
- Updated get_palantir_csv() to handle name collisions. Currently addresses this by adding the last digits of OID to the concept_set_name.

output/
- Added new outputs

data/
- cset.csv: Added a missing OID

.run/
- Added a new about.md which explains what this folder is for (JetBrains debug configs)
```


joeflack4:
@Sigfried It looks like I was able to get what Stephanie needed. I did this on my `joe` branch.

However, I haven't looked at your changes to `master` today. As I expected, there are some conflicts between our branches now. If you think that your code you wrote earlier will not be useful, we can just make the `joe` branch the new `master` branch. Or, we can merge, and either (or both of us) can handle the conflicts.

data/cset.csv
```diff
@@ -225,4 +225,5 @@
 2.16.840.1.113762.1.4.1247.71,1000000232,https://docs.google.com/spreadsheets/d/1jzGrVELQz5L4B_-DqPflPIcpBaTfJOUTrVJT5nS_j18/edit#gid=612526229,VSAC,,
 2.16.840.1.113762.1.4.1247.86,1000000233,https://docs.google.com/spreadsheets/d/1jzGrVELQz5L4B_-DqPflPIcpBaTfJOUTrVJT5nS_j18/edit#gid=612526229,VSAC,,
 2.16.840.1.113762.1.4.1247.70,1000000234,https://docs.google.com/spreadsheets/d/1jzGrVELQz5L4B_-DqPflPIcpBaTfJOUTrVJT5nS_j18/edit#gid=612526229,VSAC,,
-2.16.840.1.113762.1.4.1196.788,1000000235,https://docs.google.com/spreadsheets/d/1jzGrVELQz5L4B_-DqPflPIcpBaTfJOUTrVJT5nS_j18/edit#gid=612526229,VSAC,,
```
joeflack4, line 228: @Sigfried Hmm, interesting. I only added one OID to the end of this file. It looks like it's saying here that there are only a couple lines that are different. However, when I `git diff`, it shows that all the lines are different, and that there are `^M` (a line break character) at the end of all my lines. I guess this is something that Excel did.

Looks like this will be a pretty easy merge, then.

PR #52, opened 2022-03-15
save the enclave_codeset_id to the dataframe and write it out to the …

stephanieshong:
…csv file for us to reference.


enclave_wrangler/main.py
```diff
@@ -51,13 +51,18 @@ def run(input_csv_folder_path):
     if DEBUG:
         log_debug_info()
 
+    # Read data
+    code_sets_df = pd.read_csv(os.path.join(input_csv_folder_path, 'code_sets.csv')).fillna('')
+    # For some reason, was being read with .0's at the end.
+    code_sets_df['enclave_codeset_id'] = pd.to_numeric(code_sets_df['enclave_codeset_id'], errors='coerce')\
```
joeflack4, line 57: @stephanieshong This fixed the `.0` problem. 

Also, I noticed that some of the rows in `data/palantir-three-file/Lisa3/2022.03.02/output/code_sets.csv` previously had this problem as well. When I ran the script, it fixed all rows where this was previously an issue.

Also, you and I were confused where, when we were running the script during our meeting, it 2 rows were being updated instead of 1. From what I could ascertain, the reason that the row with `1000000181` was being updated in addition to `1000000273` (our test case) was because `1000000181` was the only one that didn't have a `.0` on it already. The other rows did; so they were not changed. I don't know if I'm explaining this very well, but either way, it is fixed now.

Oh, also, I found [diffMerge](https://sourcegear.com/diffmerge/) to be useful here to look at exactly which cells were changed. I believe that @Sigfried uses this or something similar.
diff

enclave_wrangler/main.py
```diff
@@ -142,6 +147,8 @@ def run(input_csv_folder_path):
 
 
     # --- Steph: check the failed cases 2/15/22-----------------------------
+    # @Steph: Can we remove this block yet? I don't mind keeping for a while for convenienice, but I like the practice
```
joeflack4, line 150: Usually what I do if I want to save my test code is, I add a folder to `.gitignore`. I often call mine `_archive/`. And inside of that, I put personal notes, test files, etc, relating to the project; things I don't want to commit. In cases where I have some debugging / test code that I want to keep, I put a copy of it in there.

enclave_wrangler/main.py
```diff
@@ -155,12 +162,14 @@ def run(input_csv_folder_path):
...
     for premade_codeset_id in premade_codeset_ids:
-        #    if premade_codeset_id != 1000000058:
-        #        continue
+        # TODO: temporary debug code to look for missing concept container not showing in the UI
+        # TODO: debug code for adding expressionItems to missing container from UI, l162,l163
+        if premade_codeset_id != temp_testing_cset_id:
+            continue
 
         # Do a test first using 'validate'
-        api_url = API_VALIDATE_URL
```
joeflack4, line 163: This wasn't needed anymore.

enclave_wrangler/main.py
```diff
@@ -203,7 +220,7 @@ def run(input_csv_folder_path):
         print('premade_codeset_id: ' + str(premade_codeset_id))
```
joeflack4, line 220: Not important for the near future, but eventually we might want to use the Python `logging` module to usefully handle / toggle on/off info, warnings, and error messages.

enclave_wrangler/main.py
```diff
@@ -217,9 +234,34 @@ def run(input_csv_folder_path):
...
+    # date_str = datetime.now().strftime('%Y_%m_%d_%H_%M')
+    # output_filename = 'code_sets_updated_' + date_str + '.csv'
+    output_filename = 'code_sets.csv'
+    code_sets_df.to_csv(os.path.join(input_csv_folder_path, output_filename), index=True, encoding='utf-8')
+
+    # TODO:
+    persistence_csv_path = os.path.join(PROJECT_ROOT, 'data', 'cset.csv')
+    code_sets_df_limited = code_sets_df[['enclave_codeset_id', 'enclave_codeset_id_updated_at', 'concept_set_name']]
+    persistence_df = pd.read_csv(persistence_csv_path).fillna('')
+    persistence_df_new = persistence_df.merge(
```
joeflack4, line 261: @Sigfried Stephanie wrote some code to save some of the assigned enclave IDs in a specific file. I updated this code to also update the file that we're currently mainly using for persistence, `cset.csv`.

joeflack4: @Sigfried Oh, also I saw that the columns `grouped_by_bids` and `concept_id` were empty in `cset.csv`, and I don't think they were being used, so I deleted them.

enclave_wrangler/main.py
```diff
@@ -155,12 +162,14 @@ def run(input_csv_folder_path):
     # II. call the REST APIs to create them on the Enclave
     # ...now that we have all the data from concept set are created
     # problem with 16(3623) and 39(5104) 58(1820) 73(6791) 74(1501)
+    temp_testing_cset_id = 1000000273  # Stephanie said this was a draft or archived set - Joe 2022/03/15
     for premade_codeset_id in premade_codeset_ids:
-        #    if premade_codeset_id != 1000000058:
-        #        continue
+        # TODO: temporary debug code to look for missing concept container not showing in the UI
+        # TODO: debug code for adding expressionItems to missing container from UI, l162,l163
+        if premade_codeset_id != temp_testing_cset_id:
```
joeflack4, line 169: @stephanieshong I left these two lines here just. So right now it's only uploading a single codeset. I think we should probably change it back now?

@Sigfried Stephanie had this here where it is only uploading a single set which we know we're OK to modify (I think it is a draft or an archive set), just to help with development / testing purposes. But if we're actually going to use the `enclave_wrangler`, we'll need to comment out / remove these two lines.

joeflack4 (approved):
@Sigfried Merged. I know that you're going to use this soon. Please check my other comments addressed to you here.

enclave_wrangler/main.py
stephanieshong: the other rows would have been empty, since the code to add the enclave_codset_id value is newly added code. Therefore, only the rows where I was testing the archived concept would have contained the enclave_codeset_id value. This is really strange and does not explain why this would be the case

enclave_wrangler/main.py
stephanieshong: Thank you. Once we are done with trouble shooting with Lisa we can remove them. But until then I will leave them here.

enclave_wrangler/main.py
joeflack4: Hmm... Maybe you're right when it comes to when you started your branch, but this diff is comparing the last commit you gave me on your branch, with the commit I made after I ran it 1 time. Either way, I suppose as long as we make sure the end result is good, that's what matters most.

PR #100, opened 2022-08-24
Route: Concept set names

joeflack4:
Updates
```
TermHub API
- Add: Route: concept-set-names, with alias: 'datasets/csets/names'; routes to: /datasets/csets?fields=name
- Add: util: jq_wrapper: For calling JQ from the shell
```

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -1,15 +1,21 @@
 """TermHub backend
-https://github.com/tiangolo/fastapi"""
+
+Resources
+- https://github.com/tiangolo/fastapi
+- jq docs: https://stedolan.github.io/jq/manual/
+- jq python api docs: https://github.com/mwilliamson/jq.py
+"""
+import json
 import os.path
-import subprocess
 from pathlib import Path
-from typing import Any, Dict, List
+from subprocess import PIPE, Popen, call as sp_call
+from typing import Any, Dict, List, Union
 
 import numpy as np
 import pandas as pd
 import requests
 import uvicorn
-from fastapi import FastAPI
+from fastapi import FastAPI, Query
 from fastapi.middleware.cors import CORSMiddleware
 from pydantic import BaseModel
 
@@ -17,6 +23,7 @@ from enclave_wrangler.config import config
 
 
 PROJECT_DIR = Path(os.path.dirname(__file__)).parent
+CSETS_JSON_PATH = f'{PROJECT_DIR}/termhub-csets/temp/objects/OMOPConceptSet/latest.json'
 app = FastAPI()
 app.add_middleware(
     CORSMiddleware,
@@ -25,11 +32,18 @@ app.add_middleware(
     allow_headers=['*']
 )
 
-# jq docs: https://stedolan.github.io/jq/manual/
-# jq python api doc: https://github.com/mwilliamson/jq.py
-CSETS_JSON = f'{PROJECT_DIR}/termhub-csets/temp/objects/OMOPConceptSet/latest.json'
+
+# Utils
+def jq_wrapper(query: str):
+    """Shim around Python->Shell for calling JQ. Useful until/if we change to the JQ Python API."""
+    p = Popen(query, stdin=PIPE, stdout=PIPE, stderr=PIPE, shell=True)
+    output, err = p.communicate()
+    output = output.decode("utf-8")
+    err = err.decode("utf-8")
+    return output, err
 
 
+# Routes
 @app.get("/")
 def read_root():
     """Root route"""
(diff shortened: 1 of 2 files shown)


backend/app.py
```diff
@@ -1,22 +1,29 @@
 """TermHub backend
-https://github.com/tiangolo/fastapi"""
+
+Resources
```
joeflack4, line 3: FYI: Added list of resources we probably want to remember / look up to the top of file.

backend/app.py
```diff
@@ -25,11 +32,18 @@
     allow_headers=['*']
 )
 
-# jq docs: https://stedolan.github.io/jq/manual/
-# jq python api doc: https://github.com/mwilliamson/jq.py
-CSETS_JSON = f'{PROJECT_DIR}/termhub-csets/temp/objects/OMOPConceptSet/latest.json'
+
+# Utils
+def jq_wrapper(query: str):
```
joeflack4, line 37: @Sigfried Figured it would be good to re-use this block of code which we'll probably need each time we want to call JQ from the shel.

It was more difficult to get the shell to work with JQ in this instance. It seemed to have something to do with `'` present in the `query` that I was trying to call. I tried escaping and a bunch of other stuff, I couldn't get my typical, simple `subprocess.call(query.split(' '))` one-liner to work. I ended up needing these several lines to open up a process, grab the output from that process, and decode the bytes.

backend/app.py
```diff
@@ -172,6 +177,47 @@ class CsetsUpdate(BaseModel):
     row_index_data_map: Dict[int, Dict[str, Any]] = {}
 
 
+@app.get("/concept-set-names")
+@app.get("/datasets/csets/names")
+@app.get("/jq-cset-names")
+def cset_names() -> Union[Dict, List]:
```
joeflack4, line 183: @Sigfried I added 3 different names for this route. I checked, and all of them work:

```py
@app.get("/concept-set-names")
@app.get("/datasets/csets/names")
@app.get("/jq-cset-names")
```

These are all actually aliases. They do a call to: `csets_read(field_filter=['conceptSetNameOMOP'])`, which does the work.

If we want to keep these aliases (which I think is nice, honestly), my preference is for `concept-set-names`, but there's not much harm in keeping these other ones. It will bloat what's returned from `/`, though. We probably don't want the user to see all these aliases if we only intend for them to use one of them. But we can always filter these out manually from the response for `/` if we want.

backend/app.py
```diff
@@ -172,6 +177,47 @@ class CsetsUpdate(BaseModel):
...
+    """Get concept set names"""
+    return csets_read(field_filter=['conceptSetNameOMOP'])
+
+
+@app.get("/datasets/csets")
+def csets_read(field_filter: Union[List[str], None] = Query(default=None)) -> Union[Dict, List]:
```
joeflack4, line 189: This is the function that does the work of the alias routes:
```py
@app.get("/concept-set-names")
@app.get("/datasets/csets/names")
@app.get("/jq-cset-names")
```

If you want to use this route directly to get concept set names, the following will work: http://localhost:8000/datasets/csets?field_filter=conceptSetNameOMOP

If we want to add multiple filters, you would do it like this:
http://localhost:8000/datasets/csets?field_filter=FIELD1&field_filter=FIELD2

Previously, I've seen (and liked) this approach instead for having a param as a list:
http://localhost:8000/datasets/csets?field_filter=FIELD1,FIELD2
We could do it this way by splitting on `,` to get a list, but it is not the canonical way to do it in FastAPI and doesn't have the native support.

I haven't thought about what this might look like if we wanted do operations on nested fields, but we don't have to supply all of that JQ flexibility / functionality to the user via URL query params. Would be nice, though.

backend/app.py
```diff
@@ -172,6 +177,47 @@ class CsetsUpdate(BaseModel):
...
+    """Get concept sets
+
+    field_filter: If present, the data returned will only contain these fields. Example: Passing `conceptSetNameOMOP` as
+    the only field_filter for OMOPConceptSet will return a string list of concept set names.
+
+    Resources: jq docs: https://stedolan.github.io/jq/manual/ , jq python api doc: https://github.com/mwilliamson/jq.py
+    """
+    if field_filter:
+        if len(field_filter) > 1:
+            d = {'error': 'Currently only 1 field_filter is allowed.'}
```
joeflack4, line 199: @Sigfried Not that we need more than one `field_filter` right now; we just need `field_filter= conceptSetNameOMOP` for our current purposes. But in case a user passes multiple `field_filter`s, I return this error.

backend/app.py
```diff
@@ -172,6 +177,47 @@ class CsetsUpdate(BaseModel):
...
+            output, err = jq_wrapper(query)
+            if err:
+                return {'error': str('error')}
+            d = output.split('\n')
+            # x[1:-1]: What gets returned is a \n-delimited string of names, where each name is formatted
+            # as '"{NAME}"',so we need to remove the extra set of quotations.
+            # TODO: This operation likely needs to be done in a variety of cases, but I don't know JQ well enough to
+            #  anticipate all of the situations where this might arise. - joeflack4 2022/08/24
+            # todo: Is this List[str] preferable to List[Dict,? e.g. [{'conceptSetNameOMOP': 'HEART FAILURE'}, ...]?
+            d = [x[1:-1] for x in d]
```
joeflack4, line 213: FYI:
> `# x[1:-1]: What gets returned is a \n-delimited string of names, where each name is formatted as '"{NAME}"',so we need to remove the extra set of quotations.`
> `# TODO: This operation likely needs to be done in a variety of cases, but I don't know JQ well enough to anticipate all of the situations where this might arise. - joeflack4 2022/08/24`

backend/app.py
```diff
@@ -172,6 +177,47 @@ class CsetsUpdate(BaseModel):
...
+
+    Resources: jq docs: https://stedolan.github.io/jq/manual/ , jq python api doc: https://github.com/mwilliamson/jq.py
+    """
+    if field_filter:
+        if len(field_filter) > 1:
+            d = {'error': 'Currently only 1 field_filter is allowed.'}
+        else:
+            # TODO: Need to replace this with Python API. Otherwise, deployment will be harder and must global
+            #  installation of JQ.
+            query = f"jq '.[] | .{field_filter[0]}' {CSETS_JSON_PATH}"
```
joeflack4, line 203: @Sigfried This query returns a `\n` delimited string of quoted concept set names. However, I added this `todo` for your consideration:

> `# todo: Is this List[str] preferable to List[Dict,? e.g. [{'conceptSetNameOMOP': 'HEART FAILURE'}, ...]?`

PR #103, opened 2022-08-30
Feature: Concept set comparison

joeflack4:
Updates
```
Feature: Concept set comparison
- Update: Added table
```

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -26,7 +26,7 @@ import jq
 
 DEBUG = True
 PROJECT_DIR = Path(os.path.dirname(__file__)).parent
-JSON_PATH = f'{PROJECT_DIR}/termhub-csets/temp/objects'
+JSON_PATH = f'{PROJECT_DIR}/termhub-csets/objects'
 CSETS_JSON_PATH = f'{JSON_PATH}/OMOPConceptSet/latest.json'
 
 app = FastAPI()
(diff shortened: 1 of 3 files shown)


frontend/src/CSets.js
```diff
@@ -114,19 +114,24 @@ function CSetsFromDisk(props) {
...
-          .then((res) => res.data.data.map(d => d.properties))
-  );
+          .then((res) => res.data.map(d => d.properties))
+  )
   if (isLoading) return "Loading...";
 
-  if (error) return "An error has occurred: " + error.message;
+  if (error) {
+    const err = "An error has occurred: " + error.message;
+    console.error(err)
```
joeflack4, line 132: @Sigfried what you had looked really aesthetic, but did not provide the interactivity provided by `console.error` (and also probably from `console.log()`). There's also `console.debug()` and a few others; I haven't really used these before but I should try to play w/ them.

frontend/src/CSets.js
```diff
@@ -114,19 +114,24 @@ function CSetsFromDisk(props) {
...
-  let path = 'objects/OMOPConceptSet';
-  let url = enclave_url(path)
+  let url = url_concept_sets
   let navigate = useNavigate();
   const { isLoading, error, data, isFetching } = useQuery([url], () =>
       axios
           .get(url)
-          .then((res) => res.data.data.map(d => d.properties))
-  );
+          .then((res) => res.data.map(d => d.properties))
```
joeflack4, line 126: The source of the 'map undefined' error. I had a feeling it'd be something like this. If you want, we can add a try/except. I wouldn't be surprised if we get `data.data` in the future. Note that we did get this error both when fetching from (i) enclave directly, and (ii) also from the backend route that reads from disk.

PR #106, opened 2022-09-08
Concept set comparison page: MVP

joeflack4:
Updates
```
    TermHub Frontend
    - Update: Misc: (i) fixed warnings related to some CSS properties, (ii) commented out some stuff, (iii) removed some comments, (iii) renamed some routes / classes
    - Add: Components: (i) ComparisonTable

    TermHub Backend
    - Update: Misc: (i) codestyle changes (capitalization of globals / constants), (ii) removed some comments, (iii) reorganized imports

    Enclave Wrangler
    - Update: datasets.py: Now creates any missing directories for outpath if they don't exist.

    Misc
    - Update: .run/ configs
    - Update: requirements*.txt: pandasql
```


enclave_wrangler/datasets
```diff
@@ -113,6 +113,7 @@ def download_and_combine_dataset_parts(datasetRid: str, file_parts: [str], outpa
         combine_parquet_files(parquet_dir, combined_parquet_fname)
         df = pd.read_parquet(combined_parquet_fname)
         if outpath:
+            os.makedirs(os.path.dirname(outpath))
```
joeflack4, line 116: @Sigfried Fixed the dir issue.

backend/app.py
```diff
@@ -6,62 +6,63 @@
 - jq python api docs: https://github.com/mwilliamson/jq.py
 """
 import json
-import os.path  # TODO: Siggie confused: are you supposed to use path.<method>(...) or os.path.<method>(...)?
```
joeflack4, line 9: @Sigfried I mentioned this once before but IDK if you saw. I always thought this was strange before. The typical pattern is `from os import path`, and then I would be doing like `my_path = path.join(...)`. But I've gotten warnings before. Idk, maybe this is solved in a newer version of Python. I should double check.

joeflack4: Hmm I got no warning in my IDE. maybe I should change it all to `from os import path`

.run/backend.run.xml
```diff
@@ -1,7 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="backend" type="Python.FastAPI">
     <option name="file" value="$PROJECT_DIR$/backend/app.py" />
-    <module name="TermHub" />
+    <module name="ValueSet-Tools" />
```
joeflack4, line 4: @Sigfried It keeps doing this. I need to do something on my end to fix

backend/app.py
```diff
@@ -6,62 +6,63 @@
...
 # load big files!
 # TODO: this is too slow for development where the backend has to restart all the time
-#       @Joe: would it be too crazy to run two backend servers, one to hold the data
-#             and one to service requests and handle logic? probably....
-ds = { name: pd.read_csv(os.path.join(CSV_PATH, name + '.csv')) for name in FAVORITE_DATASETS}
-print(ds.keys())
-
-concept = ds['concept']
-cnts = pysqldf("""
+#  @Joe: would it be too crazy to run 2 backend servers, 1 to hold the data and 1 to service requests / logic? probably.
```
joeflack4, line 39: @Sigfried 
```
#  @Joe: would it be too crazy to run 2 backend servers, 1 to hold the data and 1 to service requests / logic? probably.
#  @Siggie: Not a bad idea. If we invest time in that, may be better to do RDBMS instead, but I think running 2
#   *might* not take too much time... actually maybe it would. Need to pass data between the processes. RCP? We could
#   do it over REST, but idk. Maybe worth looking into / trying for an hour.
```

backend/app.py
```diff
@@ -6,62 +6,63 @@
...
-        FROM concept
+        FROM CONCEPT
         GROUP BY 1""")
+except FileNotFoundError:
+    print('Datasets not loaded.')
 
-app = FastAPI()
-app.add_middleware(
+APP = FastAPI()
+APP.add_middleware(
```
joeflack4, line 58: might be willing to make an exception on capitalization for this one because it doesn't look as nice constantly seeing `APP.route(...)`

backend/app.py
```diff
@@ -6,62 +6,63 @@
...
-
-concept = ds['concept']
-cnts = pysqldf("""
+#  @Joe: would it be too crazy to run 2 backend servers, 1 to hold the data and 1 to service requests / logic? probably.
+#  @Siggie: Not a bad idea. If we invest time in that, may be better to do RDBMS instead, but I think running 2
+#   *might* not take too much time... actually maybe it would. Need to pass data between the processes. RCP? We could
+#   do it over REST, but idk. Maybe worth looking into / trying for an hour.
+
+# TODO: #2: remove try/except when download datasets
+try:
```
joeflack4, line 45: @Sigfried 
> `# TODO: #2: remove try/except when download datasets`

This is temporary. Still on the train right now. I have internet but downloading the sets is a bit much.

frontend/src/CSets.js
```diff
@@ -33,11 +33,85 @@ const API_ROOT = 'http://127.0.0.1:8000'
 const enclave_url = path => `${API_ROOT}/passthru?path=${path}`
 const backend_url = path => `${API_ROOT}/${path}`
 
+
+//TODO: How to get hierarchy data?
```
joeflack4, line 37: I just reorganized the order of components here

frontend/src/CSets.js
```diff
@@ -223,114 +298,82 @@ function ConceptSetsPage(props) {
...
-              {concept.conceptId}: {concept.conceptName}
-            </ListItem>
-          })}
-        </List>
-      </div>)
-  )
-}
+// todo: move this back in, or accept cols as params
+// https://www.ag-grid.com/react-data-grid/column-definitions/
+const getColumnDefs = () => {
```
joeflack4, line 303: > `// todo: move this back in (to the CsetComparisonPage component), or accept cols as params`

frontend/src/CSets.js
```diff
@@ -223,114 +298,82 @@ function ConceptSetsPage(props) {
...
+  return [
+    { field: 'ConceptID' },
+    // { field: '23007370' },
+    // { field: '23600781' },
+  ];
+};
 
-function ConceptList(props) {
-  // http://127.0.0.1:8000/fields-from-objlist?objtype=OmopConceptSetVersionItem&filter=codesetId:822173787|74555844
+// TODO: Page state refresh should not be 'on window focus', but on autocomplete widget selection
```
joeflack4, line 311: @Sigfried eventually, ideally`// TODO: Page state refresh should not be 'on window focus', but on autocomplete widget selection`

frontend/src/CSets.js
```diff
@@ -223,114 +298,82 @@ function ConceptSetsPage(props) {
...
+// TODO: Should we move comparison logic python side and do query at new backend_url?
```
joeflack4, line 312: @Sigfried `// TODO: Should we move comparison logic python side and do query at new backend_url?`

I don't want to rely on Python too much. It's just a bit easier for me two wrangle things. But this wasn't so bad.

frontend/src/CSets.js
```diff
@@ -223,114 +298,82 @@ function ConceptSetsPage(props) {
...
+function CsetComparisonPage(props) {
```
joeflack4, line 313: New component `CsetComparisonPage`

frontend/src/CSets.js
```diff
@@ -223,114 +298,82 @@ function ConceptSetsPage(props) {
...
-  let {conceptId} = params;
-  let path = `objects/OMOPConceptSet/${conceptId}/links/omopconcepts`;
-  let url = enclave_url(path)
-  const { isLoading, error, data, isFetching } = useQuery([path], () =>
-      axios
-          .get(url)
-          .then((res) => res.data.data.map(d => d.properties)) )
-  */
-}
+    return axios.get(url).then((res) => {
```
joeflack4, line 319: FYI: logic for getting a data structure that shows concept set overlap / inclusion. Looks like:
`[{ConceptID: <conceptId>, <csetId1>: <'O' or 'X'>, ... <csetIdN>: <'O' or 'X'>}]`

frontend/src/CSets.js
```diff
@@ -223,114 +298,82 @@ function ConceptSetsPage(props) {
...
+        const objectOrder = {'ConceptID': null}
+        newRow = Object.assign(objectOrder, newRow);
+        tableData2.push(newRow);
+      }
+      
+      return tableData2
+  })});
 
-export {ConceptSetsPage, CsetSearch, ConceptList};
+  // TODO: How to pass column order? could I just put 'ConceptID' as the only column thta mtters
```
joeflack4, line 353: @Sigfried I'd really like to figure this out:
> `  // TODO: How to pass column order? could I just put 'ConceptID' as the only column that matters`

I visited here https://www.ag-grid.com/react-data-grid/column-definitions/ , and I saw the examples, but when I tried to replicate, it didn't work. 'ConceptID' still keeps getting put last. but it needs to be first.

### edit
i think i figured it out. i'm passing the column def to the AGGrid wrapper component I made, not the actual AGGrid component

joeflack4: Also, for the columns, I think leaving the IDs of csets as the column header is nice, but it might be nice to have some way to know the name. Maybe a mouseover, or a 'key' for the header meanings somewhere.

frontend/src/Table.js
```diff
@@ -98,4 +98,41 @@ const Table = (props) => {
   );
 };
 
-export default Table;
\ No newline at end of file
+// TODO's: (i) finish this implementation: "x"s -> checkmarks?, (ii) matrix widget too/instead? (header x side header)
+const ComparisonTable = (props) => {
```
joeflack4, line 102: This is basically a simplified version of your other Table widget. It may be possible to combine these together, or maybe we can keep them separate. im not sure yet

requirements.txt
```diff
@@ -34,39 +37,3 @@ virtualenv-clone==0.5.7
 virtualenvwrapper==4.8.4
 watchfiles==0.15.0
 websockets==10.3
-
-bs4~=0.0.1
```
joeflack4, line 38: With all these deletes, I'm guessing that one or both of us accidentally pip freezed while virtual env not active, maybe.

frontend/src/index.js
```diff
@@ -21,8 +21,8 @@ root.render(
 
         <Route path="/" element={<App />}>
           {/*<Route path="ontocall" element={<EnclaveOntoAPI />} />*/}
-          <Route path="csets-from-disk" element={<CsetSearch/>} />
```
joeflack4, line 24: renamed this route url 'cset-comparison'

joeflack4:
Screen Shot 2022-09-08 at 10 09 03 PM

frontend/src/CSets.js
Sigfried: Yeah. Refresh on window focus is annoying.

PR #124, opened 2022-09-23
Datasets download updates

joeflack4:
Updates
```
WIP:
- Update: remove "Unnamed: x" columns from datasets
- Update: move jupyter transform code over to enclave_wrangler
- Update: datasets: filter concept sets that have no container
```

### DIFF
--- a/.run/backend.run.xml
+++ b/.run/backend.run.xml
@@ -2,7 +2,7 @@
   <configuration default="false" name="backend" type="Python.FastAPI">
     <option name="additionalOptions" />
     <option name="file" value="$PROJECT_DIR$/backend/app.py" />
-    <module name="TermHub" />
+    <module name="ValueSet-Tools" />
     <option name="INTERPRETER_OPTIONS" value="" />
     <option name="PARENT_ENVS" value="true" />
     <option name="SDK_HOME" value="" />
--- a/enclave_wrangler/config.py
+++ b/enclave_wrangler/config.py
@@ -1,5 +1,7 @@
 """Config"""
 import os
+from collections import OrderedDict
+
 from dotenv import load_dotenv
 
 
@@ -19,16 +21,21 @@ config = {
     'ONTOLOGY_RID': os.getenv('ONTOLOGY_RID', 'ri.ontology.main.ontology.00000000-0000-0000-0000-000000000000'),
 }
 
-FAVORITE_DATASETS = {
-    'concept': {
+# Ordered because of transformation dependencies
+FAVORITE_DATASETS = OrderedDict({
+    'concept_set_members': {
+        'name': 'concept_set_members',
+        'rid': 'ri.foundry.main.dataset.e670c5ad-42ca-46a2-ae55-e917e3e161b6',
+    },
+    'concept': {  # transform depends on: concept_set_members transform
         'name': 'concept',
         'rid': 'ri.foundry.main.dataset.5cb3c4a3-327a-47bf-a8bf-daf0cafe6772',
     },
-    'concept_ancestor': {
+    'concept_ancestor': {  # transform depends on: concept_set_members transform
         'name': 'concept_ancestor',
         'rid': 'ri.foundry.main.dataset.c5e0521a-147e-4608-b71e-8f53bcdbe03c'
     },
-    'concept_relationship': {
+    'concept_relationship': {  # transform depends on: concept_set_members transform
         'name': 'concept_relationship',
         'rid': 'ri.foundry.main.dataset.0469a283-692e-4654-bb2e-26922aff9d71'
     },
@@ -36,17 +43,13 @@ FAVORITE_DATASETS = {
         'name': 'concept_set_container_edited',
         'rid': 'ri.foundry.main.dataset.8cb458de-6937-4f50-8ef5-2b345382dbd4',
     },
-    'code_sets': {
+    'code_sets': {  # transform depends on: concept_set_container_edited untransformed
         'name': 'code_sets',
         'rid': 'ri.foundry.main.dataset.7104f18e-b37c-419b-9755-a732bfa33b03',
     },
     'concept_set_version_item': {
         'name': 'concept_set_version_item',
         'rid': 'ri.foundry.main.dataset.1323fff5-7c7b-4915-bcde-4d5ba882c993',
-    },
-    'concept_set_members': {
-        'name': 'concept_set_members',
-        'rid': 'ri.foundry.main.dataset.e670c5ad-42ca-46a2-ae55-e917e3e161b6',
     },
     # not downloadable -- yet
     # 'safe_harbor_term_usage': {
@@ -58,4 +61,8 @@ FAVORITE_DATASETS = {
     #     'name': '',
     #     'rid': '',
     # },
+})
+FAVORITE_DATASETS_RID_NAME_MAP = {
+    v['rid']: k
+    for k, v in FAVORITE_DATASETS.items()
 }
--- a/enclave_wrangler/config.py
+++ b/enclave_wrangler/config.py
@@ -26,30 +26,37 @@ FAVORITE_DATASETS = OrderedDict({
     'concept_set_members': {
         'name': 'concept_set_members',
         'rid': 'ri.foundry.main.dataset.e670c5ad-42ca-46a2-ae55-e917e3e161b6',
+        'sort_idx': ['codeset_id', 'concept_id'],
     },
     'concept': {  # transform depends on: concept_set_members transform
         'name': 'concept',
         'rid': 'ri.foundry.main.dataset.5cb3c4a3-327a-47bf-a8bf-daf0cafe6772',
+        'sort_idx': ['concept_id'],
     },
     'concept_ancestor': {  # transform depends on: concept_set_members transform
         'name': 'concept_ancestor',
-        'rid': 'ri.foundry.main.dataset.c5e0521a-147e-4608-b71e-8f53bcdbe03c'
+        'rid': 'ri.foundry.main.dataset.c5e0521a-147e-4608-b71e-8f53bcdbe03c',
+        'sort_idx': ['ancestor_concept_id', 'descendant_concept_id'],
     },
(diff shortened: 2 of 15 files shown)


.run/backend.run.xml
```diff
@@ -2,7 +2,7 @@
   <configuration default="false" name="backend" type="Python.FastAPI">
     <option name="additionalOptions" />
     <option name="file" value="$PROJECT_DIR$/backend/app.py" />
-    <module name="TermHub" />
+    <module name="ValueSet-Tools" />
```
joeflack4, line 5: Crap these accidentally snuck in again. I'll fix this soon.

enclave_wrangler/config.py
```diff
@@ -19,35 +21,36 @@
     'ONTOLOGY_RID': os.getenv('ONTOLOGY_RID', 'ri.ontology.main.ontology.00000000-0000-0000-0000-000000000000'),
 }
 
-FAVORITE_DATASETS = {
-    'concept': {
+# Ordered because of transformation dependencies
+FAVORITE_DATASETS = OrderedDict({
```
joeflack4, line 25: `OrderedDict` because of transformation dependencies.

enclave_wrangler/config.py
```diff
@@ -58,4 +61,8 @@
     #     'name': '',
     #     'rid': '',
     # },
+})
+FAVORITE_DATASETS_RID_NAME_MAP = {
```
joeflack4, line 65: Quick lookup map to convert RID to name.

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
+#!/usr/bin/env python3
```
joeflack4, line 1: I renamed the file by adding `.py` to the end. Crazy that it wasn't there before. But now can't see diff ;(

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+from enclave_wrangler.utils import log_debug_info
+
+HEADERS = {
+    "authorization": f"Bearer {config['OTHER_TOKEN']}",
+    #"authorization": f"Bearer {config['PALANTIR_ENCLAVE_AUTHENTICATION_BEARER_TOKEN']}",
+    #'content-type': 'application/json'
+}
+DEBUG = False
+# TODO: Once git LFS set up, dl directly to datasets folder, or put these in raw/ and move csvs_repaired to datasets/
+CSV_DOWNLOAD_DIR = os.path.join(TERMHUB_CSETS_DIR, 'datasets', 'large-files-not-ready-to-push')
```
joeflack4, line 32: Updated these path variables:
```py
CSV_DOWNLOAD_DIR = os.path.join(TERMHUB_CSETS_DIR, 'datasets', 'large-files-not-ready-to-push')
CSV_TRANSFORM_DIR = os.path.join(TERMHUB_CSETS_DIR, 'datasets', 'csvs_repaired')
```

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+        df = df[df['concept_set_name'].isin(container_concept_set_name_ids)]
+    except FileNotFoundError:
+        print('Warning: Tried transforming code_sets.csv, but concept_set_container_edited.csv must be downloaded and '
+              'transformed first. Try running again after this process completes. Eventually need to do these '
+              'transformations dependency ordered fashion.', file=sys.stderr)
+
+    return df
+
+
+def transforms_common(df: pd.DataFrame) -> pd.DataFrame:
```
joeflack4, line 238: A function for common transformations. Ideally I'd read the file in here as well to avoid duplication, but one of the transformations (concept set members) involves some special operations on file read, so for now I am doing a file read in every function.

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+    """Common transformations"""
+    # - Removed 'Unnamed' columns
+    for col in [x for x in list(df.columns) if x.startswith('Unnamed')]:  # e.g. 'Unnamed: 0'
```
joeflack4, line 241: Now captures any 'Unnamed' column.

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+        df.drop(col, axis=1, inplace=True)
+    return df
+
+
+def transform(dataset_name: str):
+    """Data transformations"""
+    dataset_funcs = {  # todo: using introspection, can remove need for this if function names are consistent w/ files
```
joeflack4, line 248: How I route datasets to their appropriate transform function.

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+    parser.add_argument(
+        '-r', '--ref',
+        default='master',
+        help='Should be the branch of the dataset -- I think. Refer to API documentation at '
+                'https://unite.nih.gov/workspace/documentation/developer/api/catalog/services/CatalogService/endpoints/getTransaction')
+    parser.add_argument(
+        '-f', '--favorites',
+        default=False, action='store_true',
+        help='Just download all the datasets that are currently hardcoded as "favorites".')
+    parser.add_argument(
```
joeflack4, line 328: New param `--transforms-only`. Can work in combination with `--favorites`.

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+    df = pd.read_csv(os.path.join(CSV_DOWNLOAD_DIR, dataset_name + '.csv'), keep_default_na=False).fillna('')
+    df = transforms_common(df)
+
+    # JOIN
+    try:
+        csc_df = pd.read_csv(
+            os.path.join(CSV_TRANSFORM_DIR, 'concept_set_container_edited.csv'), keep_default_na=False).fillna('')
+        container_concept_set_name_ids = set(csc_df['concept_set_id'])
+        df = df[df['concept_set_name'].isin(container_concept_set_name_ids)]
+    except FileNotFoundError:
```
joeflack4, line 230: I left this `FileNotFoundError` warning just in case, but because I added `OrderedDict` for our favorite datasets, the datasets that are depended on should be transformed first, so we should not get these errors.

```py
        print('Warning: Tried transforming code_sets.csv, but concept_set_container_edited.csv must be downloaded and '
              'transformed first. Try running again after this process completes. Eventually need to do these '
              'transformations dependency ordered fashion.', file=sys.stderr)
```

I'm running the dataset transforms now. When they're done, assuming I got no errors, I can remove these warnings.

enclave_wrangler/datasets.py
```diff
@@ -0,0 +1,355 @@
...
+def transform_dataset__code_sets(dataset_name: str) -> pd.DataFrame:
+    """Transformations to code_sets.csv"""
+    df = pd.read_csv(os.path.join(CSV_DOWNLOAD_DIR, dataset_name + '.csv'), keep_default_na=False).fillna('')
+    df = transforms_common(df)
+
+    # JOIN
+    try:
+        csc_df = pd.read_csv(
+            os.path.join(CSV_TRANSFORM_DIR, 'concept_set_container_edited.csv'), keep_default_na=False).fillna('')
+        container_concept_set_name_ids = set(csc_df['concept_set_id'])
```
joeflack4, line 228: Here's how I did the filter operation on `code_sets.csv`:
```py
        container_concept_set_name_ids = set(csc_df['concept_set_id'])
        df = df[df['concept_set_name'].isin(container_concept_set_name_ids)]
```

enclave_wrangler/datasets.py
Sigfried: @joeflack4, with --transforms-only, will transforms happen on files that have already been transformed? or are the originals saved and transforms happen on those?

Sigfried:
Not sure if this is working. Ran it and still finding asthma_nmh in code_sets but not container:

    ➜ grep asthma_nmh termhub-csets/datasets/concept_set_container_edited.csv
    
    ➜ grep asthma_nmh termhub-csets/datasets/code_sets.csv
    2250,879212923,asthma_nmh (v1),,asthma_nmh,UNITE,1.0,2021-02-18 03:25:44+00:00,,True,1.0,,filter out asthma pt,,,v1,,,,741a70f7-7c49-4b0a-860c-a4208dc8ced7,,ri.blobster.main.configuration.8cd623c8-abcf-4101-aeb2-81ae5e860dd7,,,False


Also, might want to do something to get rid of these messages:

    /Users/sigfried/git-repos/TermHub/enclave_wrangler/datasets.py:209: DtypeWarning: Columns (4) have mixed types. Specify dtype option on import or set low_memory=False.
      df = pd.read_csv(
    /Users/sigfried/git-repos/TermHub/enclave_wrangler/datasets.py:176: DtypeWarning: Columns (3) have mixed types. Specify dtype option on import or set low_memory=False.
      csm_df = pd.read_csv(
    /Users/sigfried/git-repos/TermHub/enclave_wrangler/datasets.py:195: DtypeWarning: Columns (3) have mixed types. Specify dtype option on import or set low_memory=False.
      csm_df = pd.read_csv(

Sigfried:
The problem was with symlinks. The new code_sets.csv in csvs_repaired was fine, but datasets/code_sets.csv was the old file, not a symlink. 

Maybe, to also address my question above (with --transforms-only, will transforms happen on files that have already been transformed? or are the originals saved and transforms happen on those?), we should always download to one directory, always save transforms to one directory, and always read from that directory. No more symlinks. And we'll just leave the huge files as untracked until we get git_lfs installed (or forever).

enclave_wrangler/datasets.py
joeflack4: @Sigfried I have it reading only from the 'raw' download directory, and writing to the 'repaired' directory. Btw I think it would be nice to rename these directories like `raw/` and `transformed/` or something. Or perhaps the transformed ones can just go directly into `datasets/`.

joeflack4:
@Sigfried Ah ok, I see that you figured out what happened. I do remember you saying we were getting rid of the symlinks as well.

PR #155, opened 2022-10-08
Develop

stephanieshong:
sssom mapping samples


joeflack4:
@stephanieshong @Sigfried Hmm, I think we might want to look this over. 55 commits and 25 files changed from `main` into `develop` (not normally supposed to go in that direction) is a lot.

Sigfried:
Yeah -- that's not good. I'm not sure what happened, but I probably just made it worse. I did a bunch of work today and while trying to push it, pulled @stephanieshong's changes without inspecting them. I'm going to try reverting, since I see a button for that above.

Sigfried:
I don't know what reverting it did. I'm confused about what's going on...

I think the code in develop is all ok. When I look at @stephanieshong's [commit](https://github.com/jhu-bids/TermHub/commit/e5835061acda74828e4db75390f318bcba658f49) the changes are all changes that I made recently.

joeflack4:
Hmm, that's a bit hairy. I never noticed that **revert** button! That's handy, but indeed, I just updated my branches and it looks like some of those recent commits w/ SSSOM files etc are still there.

If the revert button doesn't work out for us, here's how I'd do it:
```sh
git checkout main
git pull  # update to get the recent pr merge
git log  # look for a previous commit before the merge
git checkout <commit_hash> 
git checkout -b main_backup
git branch -D main  # delete locally the main which had just been merged into by the PR
git branch -m main  # rename my main_backup, which is currently at the last stable commit, back to main
git push -uf origin main  # overwrite the main on GitHub with this fixed local main
```

I think we should do in this case is, when you are sure that `develop` is stable (which may already be the case), just go ahead and overwrite `main` with the `develop` branch. I suppose it doesn't matter too much yet since we don't have an app in production that people are using.

PR #163, opened 2022-10-12
(i) Git LFS initial setup, (ii) Enclave wrangler bugfixes, (iii) Backend auto-transforms

joeflack4:
Updates
```
Enclave wrangler
- Bugfix: Multiple bugfixes related to changes in the Python API for dataset download and transformation.

Backend
- Update: If some datasets haven't been transformed, will now transform them on server start rather than skipping / erroring out.

TermHub Csets
- Set up git LFS
```


joeflack4:
Could potentially merge this now, but I'm still working, so I may keep adding to this PR with more changes.

PR #188, opened 2022-12-07
DB <-- Enclave wrangler unit tests

joeflack4:
Updates
```
    Enclave Wrangler
    - Update: Existing unit tests to be proper unit tests
    - Add: Unit tests for dataset upload
```


enclave_wrangler/new_enclave_api.py
```diff
@@ -556,15 +557,15 @@ def upload_concept_set_container(
...
+    response: Response = post(api_name, d, validate_first)
     if 'errorCode' in response:
         print(response, file=sys.stderr)
         print('If above error message does not say what is wrong, it is probably the case that the `concept_set_id` '
               f'already exists. You passed: {concept_set_id}')
     return response
 
 
-def make_request(api_name: str, data: Union[List, Dict] = None, validate=False, verbose=True) -> JSON_TYPE:
+def make_request(api_name: str, data: Union[List, Dict] = None, validate=False, verbose=True) -> Response:
```
joeflack4, line 568: @Sigfried I had to change `make_request()` to return `requests.Response` instead of JSON. I need the full response for testing right now, as I'm checking the `status_code`. I should've just designed it that way in the beginning. Most of the changes in `new_enclave_api.py` are related to the the typing on this.

test/test_enclave_wrangler.py
```diff
@@ -6,37 +6,125 @@
...
-
-    # TODO: teardown
-    #  1. "apiName": "delete-omop-concept-set-version"
-    #  2. there's no endpoint to delete a concept-set-container (i think that's the one). but there is a way to archive
-    #    ...so for now, we can use the following; and Amin might add a way for us to delete a specific concept set with
-    #    ...a single known name, just for us, for our unit tests:
+# TODO use proper teardown() methods
+class TestEnclaveWrangler(unittest.TestCase):
+
+    def test_upload(self):
```
joeflack4, line 33: @Sigfried Moved `backend/standalone_api_calls/change_concepts.py` here. You can run this file right now, and it's set just to run only this test case.

test/test_enclave_wrangler.py
```diff
@@ -6,37 +6,125 @@
...
+        #       }
+        #     },
+        # TODO: I think I need to do what it suggested above regarding 'most recent version'. I think this means that my
+        #  teardown step here will involve another step to update the previous version and label it most recent.
+        # todo: ask amin?: Why do i need to declare expression items?
+        # todo: how do I get the version? is the version returned in the response json?
+        # todo: where do I pass the version? omop-concept-set param? Ask amin, is that just poorly named? if that param
+        #  is indeed for the container, then is there not a way to stipulate a version? does it just delete the most
+        #  recent version?
+        # make_request(
```
joeflack4, line 89: @Sigfried Unfortunately, I couldn't complete the teardown for this test as planned, because of some or all of these issues:
- [ ] 1. TODO: I think I need to do what it suggested above regarding 'most recent version'. I think this means that my teardown step here will involve another step to update the previous version and label it most recent.
- [ ] 2. todo: ask amin?: Why do i need to declare expression items?
- [ ] 3. todo: how do I get the version? is the version returned in the response json?
- [ ] 4. todo: where do I pass the version? omop-concept-set param? Ask amin, is that just poorly named? if that param is indeed for the container, then is there not a way to stipulate a version? does it just delete the most

test/test_backend.py
```diff
@@ -4,37 +4,33 @@
...
-    # TODO: can improve by using 'mock':
-    #  https://betterprogramming.pub/why-you-should-use-a-put-request-instead-of-a-post-request-13b593b6e67c
-    url = BACKEND_URL_BASE + 'datasets/csets'
-    response = requests.put(url=url, json={
-        'dataset_path': 'test/heroes.csv',
-        'row_index_data_map': {
-            3: {
-                'first_name': 'Spider',
-                'last_name': 'Man'
+class TestBackend(unittest.TestCase):
```
joeflack4, line 13: I changed `test_backend.py` and `test_enclave_wrangler.py` to properly use `unittest`.

joeflack4 (commented):
@Sigfried Can you review and merge? I can go over this with you if you want tomorrow afternoon, maybe at 1pm, or after 3pm, not sure.

PR #209, opened 2023-01-13
concept set editing on comparison page

Sigfried:
not ready to merge yet

- [ ] Design edit UI
- [ ] #217 

Old notes:
- comparison data table
  - some style cleanup. wanted to not have styles spread between style sheets and react in-line, but it seems unavoidable
  - working on cset editing
    - can now click on a cset heading and make it editable. that column becomes checkboxes to allow adding/removing the concept on any line.
    - showing concept set card for cset being edited - showing EditInfo with description of changes queued up. doesn't work well. checking a box twice should add and remove that concept as a change.


joeflack4:
That is pretty cool, the bot linking deployment complete.

joeflack4:
My latest update:
```
Backend
- Bugfix: hierarchy(): Some concepts were in there that weren't supposed to be
- Update: hierarchy(): Now returns list of orphans that were not found in hierarchy, likely because of vocabulary differences at the time of cset creation
- Update: cr_hierarchy(): Now also returns list of orphans
- Update: Codestyle: Cleared several PEP8 violations
```

backend/app.py
```diff
@@ -216,39 +233,63 @@ def get_cset_members_items(codeset_ids: List[int] = None, con=CON) -> List[Legac
         {'codeset_ids': codeset_ids})
 
 
-def hierarchy(codeset_ids: List[int] = None, selected_concept_ids: List[int] = None) -> Dict:
-    """Get hierarchy of concepts in selected concept sets"""
-    if not selected_concept_ids:
-        selected_concept_ids = get_concept_set_member_ids(codeset_ids, column='concept_id')
-
-    # selected_roots: List[int] = top_level_cids(selected_concept_ids)
+def get_parent_children_map(root_cids: List[int], cids: List[int], con=CON) -> Dict[int, List[int]]:
```
joeflack4, line 236: @Sigfried `new_hierarchy()` -> `hierarchy_helper()` -> now called `get_parent_children_map()`

To solve our problem with `IN (ARRAY[...])`, I tried some other things other than string interpolation:

```python
params = {'root_cids': root_cids, 'cids': cids}
query = text(query_str)
query = query.bindparams(bindparam('root_cids', expanding=True))
query = query.bindparams(bindparam('cids', expanding=False))
query = query.bindparams(root_cids=tuple(root_cids), cids=tuple(cids))
```

However I got the same problem. I ended up sticking with plain string interpolation.

backend/app.py
```diff
@@ -216,39 +233,63 @@ def get_cset_members_items(codeset_ids: List[int] = None, con=CON) -> List[Legac
...
+    """New hierarchy info from sql"""
+    # int(x): Prevents SQL injection by throwing error if not an integer
```
joeflack4, line 238: @Sigfried Since I did plain string interpolation on the SQL query, I added this `int(x)` bit that should throw an error and hopefully prevent that.
```python
    # int(x): Prevents SQL injection by throwing error if not an integer
    root_cids: List[int] = [int(x) for x in root_cids]
    cids: List[int] = [int(x) for x in cids]
    root_cids: str = ', '.join([str(x) for x in root_cids])
    cids: str = ', '.join([str(x) for x in cids])
```

backend/app.py
```diff
@@ -216,39 +233,63 @@ def get_cset_members_items(codeset_ids: List[int] = None, con=CON) -> List[Legac
...
     added_count: Dict[int, int] = {}
 
-    def recurse(ids):
-        """Recurse"""
+    def recurse(ids: List[int]):
+        """Recursively build hierarchy
+        Requires variables in outer scope: (i) parent_children_map, (ii) added_count
+        Side effects: (i) updates added_count"""
+        if not ids:
+            return None
```
joeflack4, line 272: @Sigfried We previously had an issue where I was returning `{}` for leaf nodes instead of `null`, and I did some weird JSON string replacement to fix that. However this solution with `return None` is cleaner.

backend/app.py
```diff
@@ -216,39 +233,63 @@ def get_cset_members_items(codeset_ids: List[int] = None, con=CON) -> List[Legac
...
+    relationships: List[Dict] = [dict(x) for x in sql_query(con, query, return_with_keys=True)]
+    direct_relationships: List[Dict] = [
+        x for x in relationships if x['min_levels_of_separation'] == 1 and x['max_levels_of_separation'] == 1]
+    parent_children_map: Dict[int, List[int]] = {}
+    for x in direct_relationships:
+        parent_children_map.setdefault(x['ancestor_concept_id'], []).append(x['descendant_concept_id'])
+    return parent_children_map
+
+
+def hierarchy(root_cids: List[int], selected_concept_ids: List[int]) -> (Dict[int, Union[Dict, None]], List[int]):
```
joeflack4, line 261: @Sigfried This function `hierarchy()` is almost the same. Most of the work to solve the issues we had was handled by your `new_hiearchy()` function (renamed to `hierarchy_helper()` earlier and now called `get_parent_children_map()`.

The most major change is that it returns the list of orphans.

PR #221, opened 2023-01-23
DB refresh using `objects` API

joeflack4:
Updates
```
Backend
- Add: Refreshing of tables using objects API (WIP)
```

### DIFF
--- a/test/test_enclave_wrangler.py
+++ b/test/test_enclave_wrangler.py
@@ -10,10 +10,10 @@ TODO's
 import os
 import sys
 import unittest
+from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Dict, List, Union
 
-import pandas as pd
 from requests import Response
 
 from enclave_wrangler.objects_api import EnclaveClient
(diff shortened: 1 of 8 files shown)


test/test_enclave_wrangler.py
```diff
@@ -161,6 +161,46 @@ def test_upload_concept_set_version(self):
         # self.assertTrue('result' in response and not response['result'] == 'VALID')
         self.assertLess(response.status_code, 400)
 
+    # todo: after completing this 'test', create func for it in backend/db and call/assert here
+    #  - what is the ultimate goal? how many tables are we refreshing?
+    def test_get_new_objects(self):
```
joeflack4, line 166: FYI: Meat of the changes are here. So far it's doing this:
- getting new csets based on datetime
- getting container data for those csets
- getting expression items data for those csets
- getting concept set members data for those csets

test/test_enclave_wrangler.py
```diff
@@ -161,6 +161,46 @@ def test_upload_concept_set_version(self):
...
+
+        # Expression items & concept set members
+        new_csets2: List[Dict] = []
+        for cset in new_csets:
+            version: int = cset['properties']['codesetId']
+            cset['expression_items'] = get_concept_set_version_expression_items(version, return_detail='full')
+            cset['members'] = get_concept_set_version_members(version, return_detail='full')
+            new_csets2.append(cset)
+
+        # 2. Database updates
```
joeflack4, line 192: Based on what this gives now, I think this is the next step, but I don't know for sure what tables to do updates/deletes on: 
```
        # TODO: (i) What tables to update after this?, (ii) anything else to be done?. Iterate over:
        #  - new_containers: * new containers, * delete & updates to existing containers
        #  - new_csets2:
        #    - the cset properties itself
        #    - expression_items: we might want to delete its previous items as well, because of possible changes
        #      does expression item ID change when you change one of its flags?
        #    - members
```

Sigfried: I'm not quite following this

PR #225, opened 2023-01-26
README modification suggestions

chrisroederucdenver:
biggest deal was noticing that the data is in submodules, and that you need to 'git submodule init' as well as 'git submodule update'

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,19 +66,41 @@ User Interactions
 
 ### Local setup
 1. Clone the repository.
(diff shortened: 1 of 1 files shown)


joeflack4:
Thanks for this, Chris. Looks like we somehow didn't include `git submodule update` in the setup steps, I'll add this. I'm going to merge this into `main` as well in addition to `develop` and add `git submodule update` too.

PR #273, opened 2023-03-03
Universal error handling

joeflack4:
Updates
```
Universal error handling
- Update: Backend: Added a way such that all routes will return error message, including stack trace.
```

### DIFF
--- a/backend/app.py
+++ b/backend/app.py
@@ -37,7 +37,7 @@ APP.add_middleware(
     CORSMiddleware,
     allow_origins=['*'],
     allow_methods=['*'],
-    allow_headers=['*']
+    allow_headers=['*'],
 )
 APP.add_middleware(GZipMiddleware, minimum_size=1000)
 CON = get_db_connection()
@@ -212,7 +212,7 @@ def get_cset_members_items(codeset_ids: List[int] = None, con=CON) -> List[Legac
         csm: false if not in concept set members
     """
     return sql_query(
-        con, f""" 
+        con, f"""
         SELECT *
         FROM cset_members_items
         WHERE codeset_id = ANY(:codeset_ids)
@@ -227,7 +227,7 @@ def get_parent_children_map(root_cids: List[int], cids: List[int], con=CON) -> D
     cids: List[int] = [int(x) for x in cids]
     # root_cids: str = ', '.join([str(x) for x in root_cids]) or 'NULL'
     # cids: str = ', '.join([str(x) for x in cids]) or 'NULL'
-    query = f""" 
+    query = f"""
         SELECT *
         FROM concept_ancestor
         WHERE ancestor_concept_id {sql_in(root_cids)}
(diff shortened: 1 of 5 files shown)


backend/app.py
```diff
@@ -212,7 +212,7 @@ def get_cset_members_items(codeset_ids: List[int] = None, con=CON) -> List[Legac
         csm: false if not in concept set members
     """
     return sql_query(
-        con, f"""
```
joeflack4, line 215: FYI: Lots of tiny whitespace like this got removed where it didn't need to be. I don't know how that happened. I would have reverted this, but no time before vacation.

backend/utils.py
```diff
@@ -83,6 +86,23 @@ def set_nested_in_dict(d: Dict, key_path: List, value: Any):
     get_nested_from_dict(d, key_path[:-1])[key_path[-1]] = value
 
 
+def return_err_with_trace(func):
```
joeflack4, line 89: #### `return_err_with_trace` decorator
@Sigfried This is what I wrote to do the universal exception handling on the backend.

Here's what this looks like.
Screen Shot 2023-03-03 at 6 55 59 PM

The two console errors you see there are expected/intended, but we can talk more about them when I get back.

You can see I've included fields `status`, `error`, and `stacktrace`. I think we should use these to display a proper error notification to the user.

frontend/src/UploadCsv.js
```diff
@@ -37,7 +37,15 @@ function UploadCsvPage(props) {
           console.log(res.data.errors)
           console.log('full response: ')
           console.log(res)
+        } else {
+          console.log('Unexpected *response* from server: ')
+          console.log(res)
         }
+      }).catch((err) => {
```
joeflack4, line 44: FYI: I updated Axios to not just use `.then()` to also not just respond to `200`'s and print their error messages (which is how I was returning them before because I didn't know how to show otherwise). I also added `.catch()` to properly handle errors as well.

backend/routes/cset_crud.py
```diff
@@ -295,8 +296,8 @@ def route_json_upload_new_container_with_concepts(d: UploadJsonNewContainerWithC
     return response  # todo: return. should include: assigned codeset_id's
 
 
-# TOOD: do the same try/except validate first false -> true fo all these upload routes
 @router.post("/upload-csv-new-cset-version-with-concepts")
+@return_err_with_trace
```
joeflack4, line 300: @Sigfried Usage of the `return_err_with_trace` decorator.

I wanted to implement this as a replacement for `router.post()`, to kind of wrap around that 'entirely' so that we would just need to use one decorator for all of our routes. But I'm not sure how easy that would be. Something to do in the future. For now, I think we'll want to (a) use this `return_err_with_trace` decorator like this wherever we want to return any errors explicitly, or we could just (b) use it on all routes for now. I guess probably (b).

PR #283, opened 2023-03-28
Testing feature using a func that raises exception

joeflack4:
## Description
copilot:summary

## More
copilot:all


joeflack4:
feature isn't launched yet, it looks like

PR #284, opened 2023-03-28
DB docs update - default user role

joeflack4:
Updated docs to set DB name and PGUSER to the same values as production, for ease of exporting/importing pg_dumps and having the same things work between local and production databases.


joeflack4:
Maybe let's not worry about this now. We can re-open this / do it when have time.

PR #297, opened 2023-04-17
Updated enclave-wrangler README

19aanand:
Reformatted schema table and updated some styling


enclave_wrangler/README.md
```diff
@@ -1,51 +1,52 @@
-## Enclave wrangler
+## **Enclave Wrangler**
```
joeflack4, line 1: Bolding in using `**` inside of a header (e.g. an H2 in this case: `##`) will not have an affect on how the markdown appears when rendered, at least not in GitHub. So we can leave the `**` out.

I can make this change after merging!

enclave_wrangler/README.md
```diff
@@ -1,51 +1,52 @@
...
 
-## Creating/Updating concept sets via CSV upload
-Example: [sample CSV](https://github.com/jhu-bids/TermHub/blob/develop/test/input/test_enclave_wrangler/test_dataset_upload/type-2-diabetes-mellitus.csv).
+## **Creating/Updating Concept Sets via CSV Upload**
+Example: [Sample CSV](https://github.com/jhu-bids/TermHub/blob/develop/test/input/test_enclave_wrangler/test_dataset_upload/type-2-diabetes-mellitus.csv).
 
-### CSV Schema
-At a minimum, all of these columns not marked 'Optional' in the description are necessary. Additional columns are allowable, but they will be ignored.
+### **CSV Schema**
+At a minimum, all of these columns not marked 'False' in the description are necessary. Additional columns are allowable, but they will be ignored.
```
joeflack4, line 8: Ahh, a required column. I was thinking about that. After merging, I think I'll change this verbiage to read something like the following instead:
> Non-required columns can can be included in the CSV, but will be ignored during processing.

joeflack4 (commented):
@19aanand Thanks so much Adit! Really appreciate this contribution.

I haven't personally used CSV upload in a while. We've been focusing on some other things. But if you are continuing to use it, I hope that it has been a smooth experience!

PR #310, opened 2023-04-28
Added `n3c.db` to `termhub-vocab` `.gitignore`

joeflack4:
#### Joe Flack --> Siggie on teams
Hmmm, the 18GB n3c.db is a bit of an issue for Git LFS. I think we may want to simply add this to the .gitignore in termhub-vocab for now and simply download it from the google drive link i linked earlier

i couldn't even do git add n3c.db without... by Joe Flack, has an attachment.

i couldn't even do git add n3c.db without my HD running out of space. My TermHub .git folder is 37GB o.O. I don't know if this is temporary, or if all files and their diffs, even Git LFS files, will be included in .git. This is indeed a lot of space

And if n3c.db works out for us we may wan... by Joe Flack

And if n3c.db works out for us we may want to delete these other files in termhub-vocab: https://github.com/jhu-bids/TermHub/pull/310/files

I experienced a lot of slowness when trying... by Joe Flack
Joe Flack
10:23 PM
Screen Shot 2023-04-27 at 10 10 58 PM

I experienced a lot of slowness when trying to do git adds / commits / pushes, etc while dealing with large files. I'd like you to merge this instead of me, just in case you encounter any trouble and want to abort: https://github.com/jhu-bids/TermHub/pull/310

It probably won't be a problem, just me being cautious. I don't know why it looks like these are new files. I only updated the .gitignore


joeflack4:
@Sigfried Can you merge this?

Sigfried:
Can I ignore this pull request or just close it or whatever? I don't want to put n3c.db in the .gitignore. I haven't added it yet because I haven't tested using it yet. Once I do, I think adding it from my machine might work.

joeflack4:
If we do add it to version control, this is going to make my hard drive very sad (as explained above). I think it will also make things harder for all of us. But I'm still open to possibly doing it. Let's discuss on Monday.

joeflack4:
Supplanted by:
- #789

PR #361, opened 2023-05-31
fixes #351

### LINKED ISSUE #351: Bug: `make counts-update` can accidentally run other `make` goals
## Overview
For now, probably best to run it manually `python3 backend/db/analysis.py --counts-update --note NOTE`. If you run `make counts-update NOTE`, it will take what you've written as space delimited goal names in addition to parsing them as a parameter. So for example if the word 'test' or 'all' appears, it will run those goals.

I should change it so that when you run this, it then uses Python to prompt the user to enter their note instead.

hugwuoke:
Prompts user to enter note if no note is specified when running make counts-update .


backend/db/analysis.py
```diff
@@ -249,7 +249,7 @@ def cli():
     if d['counts_update']:
         note = d['note'].strip() if d['note'] else None
         if not note:
-            print('Error: Must provide a --note when using --counts-update.', file=sys.stderr)
+            note = input("Please provide a note: ")
```
joeflack4, line 252: Looks like this was pretty simple!

I just noticed an issue / some more changes I'd like to request:
1. After the `if not note` block, add the following line: `note = note.strip()`. This change is semi-trivial / quality of life; will remove extra whitespace the user might accidentally enter, e.g. converting `  hello   world  ` to `hello world`.
2. Line 250 here (`note = d['note'].strip() if d['note'] else None`) can now be removed 
3. You can remove the `else` on line 253, putting `counts_update(note, d['schema'])` on the line right after what you did in '1'.
4. Can you run it and see if it allows you to just hit 'enter' (entering nothing) when it asks you for a note? I think probably it will allow that. If so, then right after line 252 (`input(...)`) , let's keep the error there. In fact, let's actually change it to an explicit error rather than a print: `raise ValueError('Must provide a note when using --counts-update.')

PR #364, opened 2023-06-01
357 db refresh button

forbeskat:


backend/app.py
```diff
@@ -384,6 +384,11 @@ def _cset_members_items(codeset_ids: Union[str, None] = Query(default=''), ) ->
     codeset_ids: List[int] = parse_codeset_ids(codeset_ids)
     return get_cset_members_items(codeset_ids)
 
+@APP.get("/db-refresh")
+def db_refresh_route():
```
joeflack4, line 388: Looks correct. I wonder if the only reason it didn't print is because when a change is made like this to the backend code, the server needs to be restarted! IDK if Siggie had a chance to go over that. It's not like the frontend unfortunately, which refreshes every time it detects a file save w/ changes.

frontend/src/pages/AboutPage.js
```diff
@@ -65,6 +66,19 @@ let LI = (props) => (
 );
 let DOCS = {};
 
+const handleClick = async () => {
```
joeflack4, line 69: This looks like a good placeholder for now.

frontend/src/pages/AboutPage.js
```diff
@@ -144,7 +158,16 @@ function AboutPage(props) {
         >
           /OMOPConceptSets?{codeset_ids.map(d => `codeset_ids=${d}`).join("&")}
         </Button>
-      <TextH2>How to make changes to a codeset (via Atlas JSON)</TextH2>
+      <TextH2>Database Refresh</TextH2>
```
joeflack4, line 161: Looks good. IDK what 'contained' variant even is, I'll take a look later and see for myself.

frontend/src/pages/AboutPage.js
```diff
@@ -435,4 +460,6 @@ export {
   TextH1,
   LI,
   howToSaveStagedChanges /*HelpWidget, HelpButton,*/,
+  axiosPost,
```
joeflack4, line 463: Hmm, I'm not sure these two exports are needed. `axiosPost` should come from importing the `axios` library here and I would imagine doesn't need to be exported. `handleClick` I don't think should be needed anywhere else either.

Actually, now that I think of it, `handleClick` is too generic of a name. Can you change it to `handleRefresh`?

joeflack4 (changes requested):

joeflack4:
Also thanks for doing this as a PR! I appreciate.

backend/app.py
forbeskat: Hi Joe, I actually did try restarting the backend but it didn't work

joeflack4: Ahh, ok, I see what happened. My bad @forbeskat , I led you down a wrong path. I saw that the `UploadCSV.js` used a function called `axiosPut` which I thought was imported from `axios`, but it actually is a function in TermHub's `State.js`. 

I was going to import `axiosPut` from `State.js`, but to make things simpler I used `axiosGet` instead. We may change this to a `POST` in the future if we change it so that only certain users can do the DB refresh, but for now a `GET` is fine.

Anyway after importing `axiosGet` from `State.js` and using that instead, it's working now!

joeflack4 (commented):
@Sigfried Minor FYI I also updated some styling / ordering of things on the Help / About page in this PR.

joeflack4 (approved):

PR #368, opened 2023-06-02
Unit testing for the subgraph function

hugwuoke:
Test both as a function and as an endpoint

### DIFF
--- a/test/test_backend.py
+++ b/test/test_backend.py
@@ -232,6 +232,36 @@ class TestBackend(unittest.TestCase):
                              )
                          ]
                          )
+        #Testing a relationship where a common ancestor is needed to connect the graph
+        edges3 = subgraph([4027255, 43530856, 4024552, 316139, 45766164, 43530961])
+        self.assertEqual(edges3,
+                         [
+                             (
+                                 "4024552",
+                                 "316139"
+                             ),
+                             (
+                                 "316139",
+                                 "43530961"
+                             ),
+                             (
+                                 "316139",
+                                 "45766164"
+                             ),
+                             (
+                                 "321588",
+                                 "4024552"
+                             ),
+                             (
+                                 "321588",
+                                 "4027255"
+                             ),
+                             (
+                                 "4027255",
+                                 "43530856"
+                             )
+                         ]
+                         )
             
 # Uncomment this and run this file directly to run all tests
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


test/test_backend.py
```diff
@@ -170,7 +170,18 @@ def test_counts_over_time(self):
...
+        "tests subgraphs"
+        ids = [1738170, 1738171, 1738202, 1738203]
+        edges = subgraph(ids)
+        #Part 1: Test that the edge list is the right size
+        self.assertEqual(len(ids)-1,len(edges))
+        #Part 2: Test the edges have the right valus
+        for i in range(len(ids)-1):
+            self.assertEqual(edges[i][0],ids[i])
+        for i in range(1,len(ids)):
+            self.assertEqual(edges[i-1][1],ids[i])
```
Sigfried, line 183: Hey @hugwuoke, this is fancier than I imagined, which was something like:

    self.assertEqual(edges, [
        [
          "1738170",
          "1738171"
        ],
        [
          "1738170",
          "1738202"
        ],
        [
          "1738170",
          "1738203"
        ]
      ])

Your approach works for this specific example where there happens to be a single root and the other three nodes are direct children of that root. But, you'll see, I just added more things that the test should be able to do, and for those, this approach won't work.

test/test_backend.py
```diff
@@ -170,7 +171,99 @@ def test_counts_over_time(self):
                     # Part 2: all other row counts should be non-zero
                     self.assertGreater(df[col][row], 0, msg=f"Table '{row}' had 0 rows in run '{col}'")
 
-
+    def test_subgraph(self):
```
joeflack4, line 174: As long as it runs and passes, looks good to me! Take a lot of space vertically like this but that is probably the most standard way to represent it in terms of code style.

test/test_backend.py
```diff
@@ -170,7 +171,99 @@ def test_counts_over_time(self):
...
+                                 "4027255",
+                                 "43530856"
+                             )
+                         ]
+                         )
+            
 # Uncomment this and run this file directly to run all tests
-# if __name__ == '__main__':
-#     unittest.main()
+if __name__ == '__main__':
```
joeflack4, line 267: Pretty sure this is on your radar to change this back to `unittest.main()` at the end, but just making note of it.

joeflack4 (approved):

PR #379, opened 2023-06-08
Issue 345 - "Click to add/remove" tooltip appears when mouse hovers over a row in CSET Search page.

mtj0712:


Sigfried:
@mtj0712. I merged it, but it doesn't work great. If you hover near the edge between rows of the related csets table, it flickers. And the tooltip disappears sometimes when moving the mouse around on the selected table (but that doesn't necessarily need to be fixed.)

Issue #330, opened 2023-05-23
(low) CSET COMPARISON: consider making column widths dynamically change with a max width for concept name col or at least allow user to adjust col width manually if desired. try to get concept name text wrapping to be hanging indent.


mtj0712:

hugwuoke:
@Sigfried I think Matthew worked on this and completed it successfully.

PR #380, opened 2023-06-09
Cset comparison col width

mtj0712:
I think I resolved issue #330. However, this branch is out of sync with the develop branch, so that must be adjusted.

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -103,6 +103,11 @@ body {
   padding: 0px;
 }
 
+.concept-name-text {
+  padding-left: 22px;
+  text-indent: -22px;
+}
+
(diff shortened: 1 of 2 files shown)


Sigfried:
This is better for now. I haven't looked at it on different sized screens. Also, I may be adding a couple additional columns soon and may want the ability for users to hide/unhide columns. So it might make sense for the column widths to be more dynamically calculated. I'm not merging in case you want to work on it more first. I did merge develop into this branch and resolved the conflicts (your changes happened in a couple places that got refactored to remove redundancy, so you should check to make sure I didn't lose anything). Or if you want to merge this into develop now and come back to this later, feel free.

Issue #390, opened 2023-06-13
Download indented csv of displayed concepts


Sigfried:
## Overview
New feature, to help SMEs in particular. We need to "print" out a hierarchy displayed on TermHub and download i as a CSV of displayed concepts. Hierarchy should be represented with indents.

## Sub-tasks
- [x] 1. Have a button that downloads a CSV of indented concepts
- [x] 2. Please move counts to after the last level, followed by vocabulary
- [x] 3. Have the counts be in the same order as in the TermHub page
- [x] 4. 3 last new empty columns: Include, Exclude, Notes
- [x] 5. Give more room to the download button

### Is your feature request related to a problem? Please describe.
"Printing" a screen or scraping loses the hierarchical structure, making the print-out no better than exporting the csv from the Enclave Concept Set Browser

### Describe the solution you'd like
A button (?) that, when pressed, exports the current screen to a csv, with the hierarchy maintained

### Describe alternatives you've considered
"Print" from the browse file menu, selected everything on the screen, copying, and pasting

mtj0712:
@Sigfried I made a pull request #393 with cset_comparison_new_info_panel branch. I tried to incorporate the requests Harold made above, but I'm not sure if I did it properly... Please review and tell me if anything should be done.

joeflack4:
Just copied over information from duplicate #377 and placed in the body of this issue. Also added Harold's 4 sub-tasks as a checkbox list in the body.

mtj0712:
This issue is solved by pull request #393. Closing the issue.

joeflack4:
Looks great guys!
<img width="1236" alt="Screen Shot 2023-06-16 at 5 10 31 PM" src="https://github.com/jhu-bids/TermHub/assets/13045020/2d600fbd-0136-41a0-917a-fa5fb2f3bf2b">

PR #393, opened 2023-06-15
Cset comparison new info panel

mtj0712:
Resolved issue #390. Please review and comment on any changes that should be made.


Sigfried:
Looking good. 

Big issue (my fault): the columns showing the different concept sets being compared and which concepts belong to each (with flags) aren't showing up. Let me know if you want help with that. It might not be that easy to do.

@hlehmann17 also wanted the column names to be the same as in the interface: same capitalization, spaces, etc. Of the column in the download that don't appear in the interface, instructions and order of appearance:
  - level: move to first column; rename to "Level"
  - concept_code: after Vocabulary; rename to "Concept code"
  - concept_class_id: rename to "Concept class"
  - domain_id: rename to "Concept domain"
  - domain: rename to "Domain tables with concept"
  - domain_cnt: rename to "Domain count"
  - standard_concept: rename to "OMOP standard"
  - invalid_reason: rename to "Invalid reason"
  - pathToRoot: don't include
  - hasChildren: don't include

mtj0712:
For some reason, "click to add/remove" pop-up only appears when the developer tool (F12) is open.

joeflack4:
That is super weird! If it continues to happen and you want help, if you could send me a screenshot of what you're clicking, I can check out the branch and help troubleshoot.

Sigfried:
Resolved column names and adding concept set columns.

PR #425, opened 2023-06-22
Initialize bugfixes

joeflack4:
Initialize bugfixes
- Bugfix: Fixed issue where even when using local database, was reading/updating 'manage' table variables in production.
- Bugfix: Added 'IF NOT EXISTS' declaration to DDL that creates indexes. Not so much a bug as it is quality of life.


backend/db/ddl-14-concepts_with_counts_ungrouped.jinja.sql
```diff
@@ -17,5 +17,5 @@ SELECT DISTINCT
 FROM {{schema}}concept c
 LEFT JOIN {{schema}}deidentified_term_usage_by_domain_clamped tu ON c.concept_id = tu.concept_id);
 
-CREATE INDEX ccu_idx1 ON {{schema}}concepts_with_counts_ungrouped(concept_id);
---CREATE INDEX ccu_idx2 ON concepts_with_counts_ungrouped(concept_id);
\ No newline at end of file
+CREATE INDEX IF NOT EXISTS ccu_idx1 ON {{schema}}concepts_with_counts_ungrouped(concept_id);
```
joeflack4, line 20: @hugwuoke Adding this will get around the error you were seeing. It was trying to create an index that already existed for some reason.

backend/db/load.py
```diff
@@ -144,43 +144,47 @@ def indexes_and_derived_tables(
...
+    # todo: Improve so that it iterates over modules & statements, rather than reaading in all modules,
+    #  and concatenating into a list of statements.
     statements = get_ddl_statements(schema=schema_name)
 
     # Determine which steps still needed
     if start_step:
         last_successful_step = start_step
     else:
-        with get_db_connection(schema='') as con2:
+        with get_db_connection(schema='', local=local) as con2:
```
joeflack4, line 155: @hugwuoke I had another bug here which may have caused your issue. It was reading and updating  certain variables from the production database rather than local. Bit mistake on my part, but at least it wasn't affecting the production database in any other way.

Issue #402, opened 2023-06-16
DB Refresh: 0 members sync issue: TermHub v ConceptSetBrowser


hlehmann17:
### Overview
Mismatch between TermHub and CS Browser

### To Reproduce
In prod:
1. go to Help/Issues page
2. Past 563193300 into the How to: Load a set of concept sets box
3. Go (back) to CSet Search
4. Load concepts (already says 1 member, 0 conceps)
5. Go to CSet Comparison: No concepts
6. In concept set browser, locate concept set with version id 563193300
7. Click OMOP Concepts. See the listing of 43 concepts ≠ 0 of TermHub
(8. Clicking, in Help/Issues, on Refresh Database, does not seem to do anything)

### What browser are you using?
macOS Chrome

### Expected behavior _(optional section)_
I should see 43 concepts. (I should also see the concept set listed, when I type in Imdevimab into the search box)

### Actual behavior _(optional section)_
As above

### Screenshots _(optional section)_
<>

### Additional context _(optional section)_
Same behavior 2 days ago, when I created the concept set. I thought that, with 2 nights to update itself, I would not see this problem.

## Related
https://github.com/jhu-bids/TermHub/issues/451

joeflack4:
@hlehmann17 So sorry about this! Your guinea piggery, err, testing, is highly appreciated!

## Source of the problem
I know what the problem is. The enclave's `concept_set_members` are not as up-to-date as its other data. When we do the refresh, it's fetching new containers, versions, and expression items, but not the members.

<img width="479" alt="Screen Shot 2023-06-16 at 8 48 28 PM" src="https://github.com/jhu-bids/TermHub/assets/13045020/94ca73ef-87c9-4e96-a046-35be55978307">

I just verified and see that the enclave does show these members are there now. However, the DB refresh thinks that it's already fetched everything needed for this concept set version, so I'll need to fix that.

## Possible solutions
I prefer in the order of: **d** (_in cases where we push from termhub_) > **c** > **b** > **a**
### a. Deny fetching when cset returns >0 expression items, but 0 members
This would leads to some poor UX though. A user would rightfully wonder why they a cset wasn't fetched that they knew to exist. Although if we went this route, we could display an error message about it.
### b. fetch it and update later
[...]

PR #427, opened 2023-06-22
Filtering 0-member code sets

joeflack4:
DB refresh / initialize
- Bugfix: Several places in the refresh and initialize scripts were not properly getting/setting DB variables from local database if needed
- Bugfix: Db counts would report on tables we didn't care about or error out if they stopped existing; namely tables that end with _new or _old and are created temporarily as part of the DB refresh process, which could be going on in the background when running counts.
- Bugfix/UX: Filtering out of fetched containers and code sets with 0 members (due usually to the fact that they are new and the enclave releases concept set members more slowly than other data through its API)

Frontend
- Update: help/about page: Changed 'how to fix' section from unordered to ordered list

### DIFF
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -379,13 +379,51 @@ def csets_and_members_to_db(con: Connection, schema: str, csets_and_members: Dic
     refresh_termhub_core_cset_derived_tables(con, schema)
 
 
-def csets_and_members_enclave_to_db(con: Connection, schema: str, since: Union[datetime, str]):
+def filter_cset_and_member_objects(csets_and_members: Dict[str, List[Dict]]) -> Dict[str, List[Dict]]:
+    """Filter filter out containers and code sets that have no members."""
+    csets_and_members2 = {
+        'OMOPConcept': csets_and_members['OMOPConcept'],
+        'OmopConceptSetVersionItem': csets_and_members['OmopConceptSetVersionItem'],
+        'OMOPConceptSetContainer': [],
+        'OMOPConceptSet': []
+    }
+    containers_with_codesets_with_members = set()
+    filtered_csets = []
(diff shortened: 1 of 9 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -379,13 +379,51 @@ def csets_and_members_to_db(con: Connection, schema: str, csets_and_members: Dic
     refresh_termhub_core_cset_derived_tables(con, schema)
 
 
-def csets_and_members_enclave_to_db(con: Connection, schema: str, since: Union[datetime, str]):
+def filter_cset_and_member_objects(csets_and_members: Dict[str, List[Dict]]) -> Dict[str, List[Dict]]:
```
joeflack4, line 382: @Sigfried The primary solution for now for #402 pending better solution.

frontend/src/pages/AboutPage.js
```diff
@@ -127,30 +127,35 @@ function AboutPage(props) {
 
       <TextH1>Database Refresh</TextH1>
       <TextBody>Will refresh the database with the latest data from the N3C Enclave.</TextBody>
+      <TextBody><b>IMPORTANT:</b> Concept set members are currently slowly updated in the API. As a tentative solution,
```
joeflack4, line 130: @Sigfried Added this text next to the refresh button.

test/test_enclave_wrangler.py
```diff
@@ -50,6 +51,22 @@ class TestEnclaveWrangler(unittest.TestCase):
     #     csets_and_members: Dict[str, List] = fetch_cset_and_member_objects(since=yesterday)
     #     # todo: what kind of assert?
 
+    def test_filter_cset_and_member_objects(self):
```
joeflack4, line 54: A new test

Issue #415, opened 2023-06-21
json download error


hlehmann17:
### Overview
A user (Michele Morris) tried downloading a json and got an internal server error:
https://termhub.azurewebsites.net//cset-download?codeset_id=933777427

joeflack4:
Same issue on dev: https://termhub-dev.azurewebsites.net/cset-download?codeset_id=933777427

PR #429, opened 2023-06-22
PR: Solving json download error #415

hugwuoke:
The first commit removes the issues with response.json(). The exception handling still needs some work because even though the response.json() line is no longer causing problems, the code below it in 183 of enclave_wrangler/utils.py is still preventing the 404 error from being handled politely.

```python
raise EnclaveWranglerErr({
            "status_code": response.status_code,
            "text": response.text,
            # ... ?
        })
```
@Sigfried I would appreciate some clarification on the purpose of the code above. Since I don't quite understand it's purpose I'm unsure how to change it to accommodate errors without JSON content. 

As a status update @joeflack4: With Siggie's help I was able to find that the reason the json download is not working is that we are calling an enclave page that does not exist. I am still trying to figure out how to correct that.

### DIFF
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -789,9 +789,10 @@ def refresh_tables_for_object():
 
 
 def get_concept_set_version_expression_items(
-    version_id: Union[str, int], return_detail=['id','full'][1], handle_paginated=False
+    version_id: Union[str, int], return_detail, handle_paginated=False
 ) -> List[Dict]:
-    """Get concept set version expression items"""
+    """Get concept set version expression items
+        return_detail required. can be 'id' or 'full' """
     version_id = str(version_id)
     items: List[Dict] = get_object_links(
         object_type='OMOPConceptSet',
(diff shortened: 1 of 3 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -789,7 +789,7 @@ def refresh_tables_for_object():
 
 
 def get_concept_set_version_expression_items(
-    version_id: Union[str, int], return_detail='full', handle_paginated=False
+    version_id: Union[str, int], return_detail=['id','full'][1], handle_paginated=False
```
hugwuoke, line 792: I messed around with this in the previous commit as well. I ended up leaving it this way, since I had changed the default I went through all the calls of this function and made sure that they would output the right version for their context.

hugwuoke:
Another note, when I was changing the response.json() error, I realized that there were a lot of places where a 404 error would raise issues that the code cannot gracefully handle. Instead of taking the time to correct all of those inconsistencies I just focused on resolving the 404 error and all those issues went away :) 

I think it would still be valuable to have proper error handling even when response.json() cannot work since the response has no JSON content. 

Another hurdle to me dealing with this issue is I'm not 100 % sure what it looks like when the exceptions are working correctly. 

But if this is something that should be worked on just let me know and I'll create an issue 👍

joeflack4:
@hugwuoke This is an awesome PR! I really appreciate you going through and explaining things in this level of detail. Sorry that I wasn't able to look at this today; I'll get back to you soon!

joeflack4:
@Sigfried Set you as the reviewer / assignee on this one. This is linked to the following issue which will close when this is merged:
- #415 
I think that this newer one that Matthew just opened may be the same issue,n ot sure:
- #440

Issue #418, opened 2023-06-21
`--counts-docs` daily/weekly aggregation


joeflack4:
## Overview
@hugwuoke Remember when we were on call earlier and we were looking at the [counts/deltas tables in the docs](https://github.com/jhu-bids/TermHub/blob/develop/docs/backend/db/analysis.md) and remarking at how they are soon going to explode due to the increasing refresh rate? I'd like you to go ahead and make some preliminary changes. We may iterate on these and possibly remove the older versions of the tables altogether and replace w/ these new ones.

## Sub-task list
- [x] 1. Once per day

### 1. Once per day
- [x] Add two new table sections, "Deltas over time, 1x/day", and "Counts over time, 1x/day".
- [x] Remove the "COMMENTS" row from these tables.
- [x] Make it so that the table shows a maximum of 1 column per day. 
  - [x] For counts, it should use the counts for the last entry that day. 
  - [x] For deltas, it should show the net total. E.g. if there were 3 runs in a day, and one of them was -10 rows, then +10, and another +10, the delta should be '20' for that day.

## Possible future tasks
I'm thinking about maybe adding some of the following. @Sigfried Your input would be appreciated if you think there's value in:
- Doing 1x/week tables in addition to / instead of 1x/day
- Comments: Keeping comments in the 1x/day 1x/week tables, but filtering out all comments for (i) automated refreshes, or (ii) users pushing the button, and only showing all comments made during that period that were "deliberate comments" from some special process. E.g. if we refresh 24 times/day and we're showing a table of weekly aggregation, and during 2 days that week, we ran a manual `--counts-update` with notes "Fixed a corrupted table" and "Added a new table", the value in the `COMMENTS` row for that given weekly column would be `"Fixed a corrupted table", "Added a new table"`.
- Line chart for counts. Either 1 line per table, or only 1 single line for the grand total. X axis for day/week. Y axis for count.
- Bar chart for deltas. X axis for day/week. Y axis for delta. Something like this:
![2D bar chart with postive and negative values each assigned its own color](https://github.com/jhu-bids/TermHub/assets/13045020/6dff0501-1b92-40b8-b11c-b1356fc366f4)

joeflack4:
Chris (**cgchute**), if you see this, I accidentally tagged you.

joeflack4:
Went with just daily aggregation now; not opting for any of the "possible future tasks" at this time.

PR #435, opened 2023-06-23
Counts_docs: Aggregate by day

hugwuoke:
This code is my general idea of what is needed, although it is not currently working. I thought I'd push now to give an update of what I've been working and perhaps when someone takes a look they'll have a better idea of how to solve the problem than I do.


backend/db/analysis.py
```diff
@@ -165,23 +165,15 @@ def counts_over_time(
...
-        for run in runs:
-            if run['timestamp'] == ts:
-                ts_dict[ts] = run['note']
-    runs_df = pd.DataFrame([ts_dict])
-    runs_df = runs_df.reindex(sorted(runs_df.columns), axis=1)
-    runs_df.index = ['COMMENT']
-    df = pd.concat([runs_df, data_df])
-
+    runs_df = pd.DataFrame([timestamps])
+    df.columns = runs_df.iloc[0].tolist()
```
hugwuoke, line 176: Basically the issue is that the timestamps have more entries than the data columns. I looked at the SQL table that is producing the timestamps and there have been a bunch of failed refreshes that have their own timestamps. I haven't been able to look at the table that produces the df (it will take me some time to figure out how to do that because I'm not too familiar with the code and SQL etc) but I think that the failed refreshes are columns that are not showing up in the df variable. This is quite interesting and we might need a different approach for adding entries to the counts_runs table, I'll have to think about this more because I think the changes I made to the variables for db_refresh is creating this issue.

joeflack4:
@hugwuoke Just got back from vacation. Thanks for this early update! I think that's a good idea. I think we both have some more important issues to work on over the next couple days at least. I should be able to get back to you about this on Friday.

backend/db/analysis.py
joeflack4: @hugwuoke Oh, I see. Good investigation! I really appreciate you taking the time to think ahead and put your reasoning here.

That definitely sounds plausible. And that sounds potentially tricky as well. I'm trying to think as well what would make the number of columns different than the number of timestamps. They should be the same, but I would not be surprised if they somehow were not either. 

For the moment, I'll leave it to you (or Siggie) to prioritize where this should go in your list of priorities. The easiest way for me to help with your priorities would be through weekly/biweekly meetings (1:1 or the whole TermHub team).

hugwuoke:
I think checking out this branches and taking a look at the new tables in analysis.md would probably be the best way to check the changes I made. But this is ready to go, let me know if you would like to transition to once a week or to have options in the function to either print a once a week view or once a day view.

test/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
joeflack4: @hugwuoke Somehow this snuck in here. Just a short 6 line file. But it's no problem. I will delete!

backend/db/analysis.py
```diff
@@ -165,43 +165,37 @@ def counts_over_time(
...
+    dates = list(set(dateslist))
+
+    finaldf = pd.DataFrame()
+    for date in dates:
+        # Create temporary table with the columns of only one date
+        datedf = df[df.columns[df.columns.str.startswith(date)]]
+        datedf = datedf.sort_index(axis=1)
+        # todo: List number of times the refresh ran that day somewhere?
+        # count = dateslist.count(date)
+        # datedf.loc[f'Number of refreshes ran'] = count
```
joeflack4, line 180: @hugwuoke I just renamed the new special rows "delta/count over time, 1x/day" to "Number of refreshes ran", but then commented it out to remove for now. Showing the number of refreshes that ran that day, which is what this was doing, may be potentially useful, but I think better to leave it off the table for now.

I was wondering why you added this here, and then I read my [initial instructions](https://github.com/jhu-bids/TermHub/issues/418#issue-1766522253) and saw that they were entirely unclear / misstated!:
>  Add two new table sections, "Deltas over time, 1x/day", and "Counts over time, 1x/day".

What I actually meant was to add new sections to the markdown file; i.e. in addition to the 2 tables in the `analysis.md` that have 1 column per run, we could have 2 more tables that aggregate these per day. But I think that what you've done is actually better; just having 1 table for counts and 1 table for deltas, and having those be aggregated to 1x/day. 

I think what we have here is good. I think no need for the weekly aggregation right now. I'll merge this and let that issue be closed. Thanks so much!

backend/db/analysis.py
```diff
@@ -272,7 +266,7 @@ def cli():
     elif d['deltas_over_time']:
         counts_over_time(method='delta_table', local=local)
     elif d['counts_docs']:
-        docs()
+        counts_docs()
```
joeflack4, line 269: Just a small change needed here.

Issue #451, opened 2023-06-28
DB refresh: Get all rows into the `concept_set_members` (and derivative) tables


Sigfried:
## Overview
Get all rows into the `concept_set_members` (and derivative) tables. Currently there are a number of concept sets that only have 1000 concepts because handle_paginated was not invoked.

Seems to also be stuck in an endless loop fetching the same thing.

## Sub-task list
- [x] 1. populate correct tables w/ all items (solve 10k response 400 err)
- [x] 2. fix endless loop (_edit_: No issue: https://github.com/jhu-bids/TermHub/pull/453#discussion_r1245963924)
- [x] 3. make sure everything in #437 discussion has been considered

### 1. populate correct tables w/ all items (solve 10k response 400 err)
Currently in #453, I've allowed it to not error out and continue the refresh. so, as a final step, in `fetch_cset_and_member_objects`, change `EnclavePaginationLimitErr` so that it does not simply accept the error and do nothing.
#### Possible solutions
a. **Datasets API dynamically during refresh, synchronously**
I think this will hold up the refresh too long and do more harm than good. Go with 'b'.
b. **Datasets API dynamically via separate GH action**
I'm thinking that we can have a GitHub action that does does the vocab tables refresh 1x/week, but it actually runs more often (e.g. 1x/30 min). When it does this, it checks if it's been a week since the last normally scheduled run and, if so, runs. Else, it checks to see if there's been any pagination errors recently (we can set a DB variable for this) and, if so, it runs the / a subset of the vocab tables refresh in order to correctly populate all the tables for the codesets that failed.
c. Later: **'a', but asynchronously in a new thread**

## Related
https://github.com/jhu-bids/TermHub/issues/402

joeflack4:
Solutions

joeflack4:
Doesn't seem like an endless loop fetching the same data. Just a really long loop. See [my comment here](https://github.com/jhu-bids/TermHub/pull/453#discussion_r1245963924).

joeflack4:
@Sigfried The DB refresh is currently not able to fetch more than 10k expression items (this can happen, but so far I've only seen it in Hope's new tests), as explained [here](https://github.com/jhu-bids/TermHub/pull/453#discussion_r1245963924). I've put in a temporary solution in this #453 (see [comment](https://github.com/jhu-bids/TermHub/pull/453#discussion_r1245955784)).

[...]

PR #453, opened 2023-06-28
DB Refresh: Syntax error bugfix && other bugfixes

joeflack4:
Updates
```
DB Refresh
- Bugfix: Syntax error causing refresh to fail.
- Bugfix: GitHub action was alwasy passing even when it should have shown a failure because exception handling was not re-raising error.
- Bugfix: Minor: Decreased excessive recently added verbosity
- Bugfix: 'None is not iterable': This was caused by unhandled exceptions due to errors being coming from the enclave after pagination handling.

Enclave Wrangler
- Update: Added todo comment to make_objects_request().
```

### DIFF
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -54,13 +54,15 @@ def refresh_db(
         counts_update('DB refresh.', schema, local)
         update_db_status_var('last_refresh_success', current_datetime(), local)
         update_db_status_var('last_refresh_result', 'success', local)
+        update_db_status_var('refresh_status', 'inactive', local)
         print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
 
     except Exception as err:
         update_db_status_var('last_refresh_result', 'error', local)
+        update_db_status_var('refresh_status', 'inactive', local)
         counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
-        print(f"Database refresh incomplete. An exception occurred: {err}")
-    update_db_status_var('refresh_status', 'inactive', local)
+        print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
+        raise err
 
 def cli():
     """Command line interface"""
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -14,6 +14,7 @@ TODO's
 """
 import json
 import os
+import sys
 from datetime import datetime
 import pytz
 from typing import List, Dict, Tuple, Union
@@ -26,7 +27,7 @@ from sqlalchemy.engine.base import Connection
 from typeguard import typechecked
 
 from enclave_wrangler.config import FAVORITE_OBJECTS, OUTDIR_OBJECTS, OUTDIR_CSET_JSON, config
-from enclave_wrangler.utils import enclave_get, enclave_post, fetch_objects_since_datetime, \
+from enclave_wrangler.utils import EnclavePaginationLimitErr, enclave_get, enclave_post, fetch_objects_since_datetime, \
     get_objects_df, get_url_from_api_path, \
     make_objects_request
 from enclave_wrangler.models import OBJECT_TYPE_TABLE_MAP, convert_row, get_field_names, field_name_mapping, pkey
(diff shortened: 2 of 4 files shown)


backend/db/refresh.py
```diff
@@ -54,13 +54,15 @@ def refresh_db(
...
         print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
 
     except Exception as err:
         update_db_status_var('last_refresh_result', 'error', local)
+        update_db_status_var('refresh_status', 'inactive', local)
         counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
-        print(f"Database refresh incomplete. An exception occurred: {err}")
-    update_db_status_var('refresh_status', 'inactive', local)
+        print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
+        raise err
```
joeflack4, line 65: Had to change a couple other lines too, but this basically fixes the issue where GitHub actions were not erroring like they were supposed to.

enclave_wrangler/objects_api.py
```diff
@@ -353,8 +356,17 @@ def fetch_cset_and_member_objects(
     for cset in cset_versions:
         version: int = cset['properties']['codesetId']
         # Hierarchical
-        cset['expression_items']: List[Dict] = get_concept_set_version_expression_items(version, return_detail='full')
-        cset['member_items']: List[Dict] = get_concept_set_version_members(version, return_detail='full')
+        # TODO: exc handling: Will need to handle differently or do more: https://github.com/jhu-bids/TermHub/issues/451
+        try:
+            cset['expression_items']: List[Dict] = get_concept_set_version_expression_items(version, return_detail='full')
+        except EnclavePaginationLimitErr as err:
```
joeflack4, line 362: @Sigfried Creating new error type and using this (maybe just temporarily) until I have a better fix in place for the pagination limit issue as it pertains to the DB refresh.

enclave_wrangler/objects_api.py
```diff
@@ -353,8 +356,17 @@ def fetch_cset_and_member_objects(
...
+            cset['expression_items']: List[Dict] = err.args[1]['results_prior_to_error']
```
joeflack4, line 363: Mind you, this is just temporary as I don't have time for a better fix today. I'd like to get your feedback on my possible solutions here:
- #451

enclave_wrangler/objects_api.py
joeflack4, line 362: Can Amin and co really not fix the `ObjectsExceededLimit` error? I can't remember why. Was this a Foundry thing?

joeflack4: Btw, this limit happened only when fetching expression items, and not during member items. I wonder if this is a coincidence, or this is just how it is w/ Enclave; that the expression items has a limit of 10k items returned, and member items do not have such a limit.

Currently I'm debugging codeset `938394329`, and so far it has 51,000 unique items. I think this may be what you were stuck on when you thought you were in an endless loop fetching the same data over and over; I don't think that's the case right now. I assume that this will eventually complete.

backend/db/refresh.py
joeflack4, line 65: Btw, duplicate index error happened: `ProgrammingError('(psycopg2.errors.DuplicateTable) relation "csmi_idx1" already exists\n')`
I haven't investigated yet, so I'm not sure yet if this error (in lieu of skipping it by adding `IF NOT EXISTS`) will be helpful or just an annoyance.
But I guess it will probably be a net positive overall.

joeflack4:
Merging this now because even though #454 is still causing issues, the current refresh running off of `develop` is broken anyway (it's just erroring silently). This code fixes a few things and gets the refresh further.

enclave_wrangler/objects_api.py
```diff
@@ -793,22 +805,23 @@ def refresh_tables_for_object():
...
+    version_id: Union[str, int], return_detail, handle_paginated=True, fail_on_error=True
 ) -> List[Dict]:
     """Get concept set version expression items"""
     version_id = str(version_id)
     items: List[Dict] = get_object_links(
         object_type='OMOPConceptSet',
         object_id=version_id,
         link_type='OmopConceptSetVersionItem',
-        handle_paginated=handle_paginated)
+        handle_paginated=handle_paginated,
```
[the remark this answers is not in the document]
Sigfried: The counts tables won't be current enough on the enclave to help, probably.

Getting the member count before getting the members is a good question -- probably for Amin. I don't know a way.

Since the new test cset versions I made yesterday and today have more than 50K members, the objects API won't work anyway -- so the test whether the object api approach and dataset api approach will yield the same results will at least fail on those two version.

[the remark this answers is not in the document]
joeflack4: Actually, I was able to fetch more than 50k members. But I wasn't able to fetch more than 10k expression items (which somehow appeared to be the case for these sets)

PR #462, opened 2023-06-30
Bugfix: DB refresh index error

joeflack4:
8ecfd6c8c20fda8954aa8cff69f676882b7b3f45
DB Refresh
- Bugfix: Indexes need to be created last. Was throwing error because the index name already exists in the same schema (on the old table not yet deleted). I am unsure why this did not cause an error before. It is possible that maybe indexes were somehow not being created, but that also seems unlikely.
- Update: Reactivated DDL for csets_to_ignore

ef4289b4bd2262cca4e7822c88a9bea96e8452db
DB Refresh
- Update: Changed method of recreating indexes to increase performance.

### DIFF
--- a/.github/workflows/refresh_db.yml
+++ b/.github/workflows/refresh_db.yml
@@ -3,8 +3,8 @@ name: DB Refresh (Objects API)
 on:
   schedule:
 #    - cron: '*/20 * * * *'  # every 20 minutes
-    - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
-#    - cron: '0 * * * *'  # every hour on the hour
+#    - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
+    - cron: '0 * * * *'  # every hour on the hour
   workflow_dispatch:
   repository_dispatch:
     types: [refresh-db]
--- a/backend/db/ddl-11-all_csets.jinja.sql
+++ b/backend/db/ddl-11-all_csets.jinja.sql
@@ -55,6 +55,6 @@ SELECT ac.*, cscnt.counts, cscnt.counts->>'Members' as concepts
 FROM ac
 LEFT JOIN {{schema}}codeset_counts cscnt ON ac.codeset_id = cscnt.codeset_id;
 
-CREATE INDEX ac_idx1 ON {{schema}}all_csets{{optional_suffix}}(codeset_id);
+CREATE INDEX ac_idx1{{optional_index_suffix}} ON {{schema}}all_csets{{optional_suffix}}(codeset_id);
 
-CREATE INDEX ac_idx2 ON {{schema}}all_csets{{optional_suffix}}(concept_set_name);
+CREATE INDEX ac_idx2{{optional_index_suffix}} ON {{schema}}all_csets{{optional_suffix}}(concept_set_name);
--- a/backend/db/ddl-12-csets_to_ignore.jinja.sql
+++ b/backend/db/ddl-12-csets_to_ignore.jinja.sql
@@ -1,5 +1,4 @@
 -- Table: csets_to_ignore ----------------------------------------------------------------------------------------------
-
 CREATE OR REPLACE VIEW {{schema}}csets_to_ignore{{optional_suffix}} AS
 SELECT all_csets.concept_set_name,
     all_csets.container_created_at,
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -5,6 +5,8 @@ todo's
 """
 import json
 import os
+from random import randint
+
 import pytz
 import dateutil.parser as dp
 from datetime import datetime, timezone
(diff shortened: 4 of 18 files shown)


backend/db/utils.py
```diff
@@ -54,24 +54,37 @@ def refresh_termhub_core_cset_derived_tables(con: Connection, schema: str):
...
+
     # Create new tables and backup old ones
+    print('Derived tables')
     t0 = datetime.now()
     for module in ddl_modules:
-        print(f'Running SQL to recreate derived table: {module}...')
+        print(f' - creating new table/view: {module}...')
         statements: List[str] = get_ddl_statements(schema, [module], temp_table_suffix, 'flat')
-        for statement in statements:
+        main_statements = [s for s in statements if not s.startswith('CREATE INDEX')]
```
joeflack4, line 64: Non-index DDL: runs first

backend/db/utils.py
```diff
@@ -54,24 +54,37 @@ def refresh_termhub_core_cset_derived_tables(con: Connection, schema: str):
...
         run_sql(con, f'DROP VIEW IF EXISTS {schema}.{view}_old;')
     for module in ddl_modules:
         run_sql(con, f'DROP TABLE IF EXISTS {schema}.{module}_old;')
+    t1 = datetime.now()
+
+    # Indexes
+    for module in ddl_modules:
+        print(f' - creating indexes for table/view: {module}...')
+        statements: List[str] = get_ddl_statements(schema, [module], temp_table_suffix, 'flat')
+        index_statements = [s for s in statements if s.startswith('CREATE INDEX')]
```
joeflack4, line 84: Index DDL: runs last, after new tables have been made and old temporarily backed up tables (and their indexes) are deleted.

backend/db/ddl-12-csets_to_ignore.jinja.sql
```diff
@@ -1,6 +1,4 @@
 -- Table: csets_to_ignore ----------------------------------------------------------------------------------------------
-
-/*
 CREATE OR REPLACE VIEW {{schema}}csets_to_ignore{{optional_suffix}} AS
```
joeflack4, line 2: @Sigfried The best option seemed to be to reactivate the DDL for `csets_to_ignore`. I couldn't delete the old version of `all_csets` because `csets_to_ignore` was dependent.

Another option is simply to delete that table. I know you told me not to worry about dealing with it last week, but I don't know if that is still the case, or if that also means I can delete it.

.github/workflows/refresh_db.yml
```diff
@@ -3,8 +3,8 @@ name: DB Refresh (Objects API)
 on:
   schedule:
 #    - cron: '*/20 * * * *'  # every 20 minutes
-    - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
-#    - cron: '0 * * * *'  # every hour on the hour
+#    - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
+    - cron: '0 * * * *'  # every hour on the hour
```
joeflack4, line 7: @Sigfried The last refresh took 40+ minutes. This is largely because of 3 new Hope test sets where it is actually able to fetch over 50k concepts (I have observed), but is still limited to 10k expressions (or maybe it's the reverse). But perhaps future runs could also take a long time, so I'm setting the refresh rate to 1x/hour instead of 1x/20min until we do some optimizations soon.

backend/db/utils.py
joeflack4: Changing this now so that we do a hash instead.

backend/db/ddl-11-all_csets.jinja.sql
```diff
@@ -55,6 +55,6 @@ SELECT ac.*, cscnt.counts, cscnt.counts->>'Members' as concepts
 FROM ac
 LEFT JOIN {{schema}}codeset_counts cscnt ON ac.codeset_id = cscnt.codeset_id;
 
-CREATE INDEX ac_idx1 ON {{schema}}all_csets{{optional_suffix}}(codeset_id);
+CREATE INDEX ac_idx1{{optional_index_suffix}} ON {{schema}}all_csets{{optional_suffix}}(codeset_id);
```
joeflack4, line 58: Changed method of re-creating indexes. Now adds a random hash suffix rather than recreating at the end.

backend/db/utils.py
```diff
@@ -48,30 +50,35 @@ def refresh_termhub_core_cset_derived_tables(con: Connection, schema: str):
...
     views = [
         'cset_members_items_plus',
-        # 'csets_to_ignore',            not using for now
+        'csets_to_ignore',
     ]
+
     # Create new tables and backup old ones
+    print('Derived tables')
     t0 = datetime.now()
+    hash_num = '_' + str(randint(10000000, 99999999))
```
joeflack4, line 63: Changed method of re-creating indexes. Now adds a random hash suffix rather than recreating at the end.

PR #465, opened 2023-07-04
Bugfix: Fetch all cset members / items

joeflack4:
## Overview
Addressing issue where, during DB refresh, we've been sometimes unable to fetch all of the cset members or version items for concept sets.

## Updates
```
    DB Refresh
    - Bugfix: Set up a system such that if the enclave errors out and doesn't allow us to fetch all concept set members or expression items, we have a way to work around that and import them.
    - Update: Refactor: Consolidated primary key code into enclave_wrangler/config.py. Also added some todo's there.
    - Add: fetch_audit table: Set up code for it in initialize(). It will be used by this bugfix and elsewhere.
    - Update: Swapped order of DDL files numbered 3 and 4.

    Enclave wrangler
    - Update: datasets.py: Now by default will re-download and overwrite whatever is in the cache / 'downloads' folder by default.
    - Bugfix: datasets.py: concept_set_vesion_item.csv was being saved with codeset_id column as float.
```


enclave_wrangler/config.py
```diff
@@ -82,7 +86,8 @@
         'rid': 'ri.foundry.main.dataset.c9932f52-8b27-4e7b-bdb1-eec79e142182',
         'sort_idx': ['concept_set_name'],
         'converters': {'archived': lambda x: True if x == 'True' else False},  # this makes it a bool field
-        'dataset_groups': ['cset']
+        'dataset_groups': ['cset'],
+        'primary_key': 'concept_set_id'
```
joeflack4, line 90: FYI: Added PKs to `enclave_wrangler/config.py`

backend/db/initialize.py
```diff
@@ -22,6 +22,35 @@
...
+note text);"""
+# fetch_audit: table schema
+#   table: text; examples: code_sets | concept_set_container | concept_set_members | concept_set_version_item
+#   primary_key: text; comma-delimited: this should store the value of the keys, e.g. codeset_id 12345, concept_id 5678
+#     would be 12345,5678
+#   status_initially: text; factor: success | fail-excessive-members | fail-0-members | fail-unknown?
+#   success_datetime: timestamp; initially null, used by code that updates the derived tables.
+#   comment: text
+# todo: change comment to text[]?
+DDL_FETCH_AUDIT = """CREATE TABLE IF NOT EXISTS public.fetch_audit (
```
joeflack4, line 48: @Sigfried I set up `initialize()` to create the `fech_audit` table, renamed from plain 'audit' for [disambiguation](https://dba.stackexchange.com/a/15187) purposes.

Some notes about the columns:
- `table` (text)
- `primary_key` (text): We initially wrote down to create two of these columns; one for the names of the primary keys in their tables, and one for the PK value(s) for a particular record. But we don't need the former, since `enclave_wrangler/[config.py](http://config.py/)` lists them now. So now it just shows values, comma-delimited, e.g. codeset_id 12345, concept_id 5678 would be: 12345,5678.
- `status_initially` (text): Renamed from 'status'. This is to disambiguate from the initial status when we first fetch, from the final status, which we can ascertain from 'success_datetime'. Right now I think it will be one of:  success | fail-excessive-members | fail-0-members | fail-unknown.
- `success_datetime` (datetime): Renamed from 'processed'. Nullable.
- `comment` (text): Left as plain text for now but can change to array if needed.

_Didn't need this column:_
- `fixed_later` (bool): Don't need this column because we can ascertain by if 'success_datetime' is not null

enclave_wrangler/objects_api.py
```diff
@@ -354,22 +352,6 @@ def fetch_cset_and_member_objects(
     flat_expression_items = []
     flat_member_items = []
     for cset in cset_versions:
-
-        # TODO: @joeflack4, need an audit log of successful as well as failed fetches so that we can make a quicker
```
joeflack4, line 358: @Sigfried RE:
> TODO: @joeflack4, need an audit log of successful as well as failed fetches so that we can make a quicker version of derived tables update

I decided not to implement _logs of successful fetches_ because I don't see how it helps with updating the derived tables:
1. We don't need to log successful fetches in order to improve derived tables (from completely remaking them to just doing inserts). `fetch_cset_and_member_objects()` returns the data and then uses it to insert directly into the core tables. So all that needs to be done is to do the same thing for the derived tables.
2. If we fetch data, log successes, but then an error occurs, the logged successes will contain any data needed to insert data anyway. We'll need to fetch the data again.

Maybe there is some other reason to log successful fetches, but I'm not clear on it. 

Is it not more important that we log _successful imports_ rather than successful fetches? That is, we should want some sort of logging done at the end of the refresh, rather than at the end of `fetch_cset_and_member_objects()`?

If I can get some clarification on of we should indeed log successful imports rather than fetches, and what that will be useful for (if only just to have some record), I will implement.

backend/db/initialize.py
joeflack4: I'm going to continue using this table schema, but upon actually trying to put it into practice, I realize that it suits more the use case you had in mind than the one that I'm immediately working on. Rather than 'table' and 'primary key', what I need to know specifically is what fetched link object failed (e.g. expression items, members) for a given code set.

enclave_wrangler/objects_api.py
```diff
@@ -351,60 +348,33 @@ def fetch_cset_and_member_objects(
...
         try:
             cset['expression_items']: List[Dict] = get_concept_set_version_expression_items(version, return_detail='full')
         except EnclavePaginationLimitErr as err:
             cset['expression_items']: List[Dict] = err.args[1]['results_prior_to_error']
-            # TODO: flag in some table that we need to do datasets api download
-            # cset['termhub-notes']['expression_items_failed_after'] = len(cset['expression_items'])
-            print(pagination_err.format(version), file=sys.stderr)
-
+            fetch_statuses.append({'table': 'code_sets', 'primary_key': version, 'status_initially':
+                'fail-excessive-items', 'comment': f"Failed after {len(cset['expression_items'])} expression_items."})
```
joeflack4, line 361: FYI: What it looks like when reporting failure to fetch enough expression items.

enclave_wrangler/objects_api.py
```diff
@@ -351,60 +348,33 @@ def fetch_cset_and_member_objects(
...
         try:
             cset['member_items']: List[Dict] = get_concept_set_version_members(version, return_detail='full')
-            auditMsg = f"found {len(cset['member_items'])} members"
         except EnclavePaginationLimitErr as err:
             cset['member_items']: List[Dict] = err.args[1]['results_prior_to_error']
-            print(pagination_err.format(version), file=sys.stderr)
-            auditMsg = f"failed after {len(cset['member_items'])} members"
-
+            fetch_statuses.append({'table': 'code_sets', 'primary_key': version, 'status_initially':
+                'fail-excessive-members', 'comment': f"Failed after {len(cset['member_items'])} members."})
```
joeflack4, line 367: FYI: What it looks like when reporting failure to fetch enough members.

enclave_wrangler/objects_api.py
```diff
@@ -351,60 +348,33 @@ def fetch_cset_and_member_objects(
...
         if not cset['member_items']:
-            auditMsg = f"got no members as of {str(datetime.now())}, will continue trying until..."
-            pass
-            # TODO: do something so we keep trying for a couple hours. it's possible it really doesn't have members, but
-            #   probably the members just aren't available yet
-
-
+            fetch_statuses.append({'table': 'code_sets', 'primary_key': version, 'status_initially': 'fail-0-members',
+                'comment': f"Fetched 0 members after fetching {len(cset['expression_items'])} items."})
```
joeflack4, line 370: FYI: What it looks like when reporting failure to fetch any members (also shows how many expression items were fetched).

test/enclave_wrangler_tests/test_utils2.py
```diff
@@ -0,0 +1,47 @@
...
+
+TEST_INPUT_DIR = os.path.join(TEST_DIR, 'input', 'test_enclave_wrangler')
+TEST_SCHEMA = 'test_n3c'
+
+
+# todo: temporarily naming this 'utils2' to make merge conflicts easier until merge Hope's new PR, then will merge this
+#  into that file.
+class TestEnclaveWranglerUtils2(unittest.TestCase):
+
+    def test_insert_fetch_status(self):
```
joeflack4, line 24: Not sure if the most useful unit test in the world, but I did verify in an easy way that things were working correctly, so probably good to have this.

enclave_wrangler/objects_api.py
```diff
@@ -295,15 +296,14 @@ def csets_and_members_enclave_to_db(
...
-
     print(f'  - Fetched new data in {(datetime.now() - t0).seconds} seconds:\n    OBJECT_TYPE: COUNT\n' +
           "\n".join(['    ' + str(k) + ": " + str(len(v)) for k, v in csets_and_members.items()]))
     return csets_and_members_to_db(con, schema, csets_and_members)
 
 
 def fetch_cset_and_member_objects(
-    since: Union[datetime, str], verbose=False
-) -> Dict[str, List[Dict]] :
+    since: Union[datetime, str] = '', cset_versions: List[Dict] = [], verbose=False
```
joeflack4, line 305: @hugwuoke @Sigfried I updated `fetch_cset_and_member_objects()` so that you can either pass `since` or a list of cset version objects (I could have also went with allowing for a list of codeset_ids instead). I think there are a variety of cases where this will be useful. Where this will most immediately be useful for me is with unit testing. I'm going to write a test where I pass it a concept set that I know is going to fail because it has more members/items than the enclave will allow us to fetch. Then it will execute the rest of the code I'm writing to do our workaround to fetch that data.

test/backend_tests/test_utils.py
```diff
@@ -0,0 +1,46 @@
...
+for codeset_id in [1, 2, 3]:
+    MOCK_FAILED_FETCH_OBJS.append({'table': 'code_sets', 'primary_key': codeset_id, 'status_initially':
+        'fail-excessive-items', 'comment': 'Unit testing.'})
+    MOCK_FAILED_FETCH_OBJS.append({'table': 'code_sets', 'primary_key': codeset_id, 'status_initially':
+        'fail-excessive-members', 'comment': 'Unit testing.'})
+    MOCK_FAILED_FETCH_OBJS.append({'table': 'code_sets', 'primary_key': codeset_id, 'status_initially':
+        'fail-0-members', 'comment': 'Unit testing.'})
+
+
+class TestBackendUtils(unittest.TestCase):
```
joeflack4, line 26: Created new test file for testing `backend/db/utils.py` and added some tests for the new `fetch_audit` table.

joeflack4: @hugwuoke Can you take a look at this? I just want you to be aware with the style I went with for testing this new code that I was working on. I learned a little in the process, kind of a new pattern.

First, I did what we're starting to do with unit tests now where we have a different `test_` file dedicated to its corresponding file in the source code. Then, I made the class name... which is kinda verbose... Idk if this is the best way to do it: test class `TestBackendDbUtilsFetchAudit`. The reason I put `FetchAudit` at the end is just for further grouping. These are related functions. And also they have a common setup/teardown step, where before any tests in the class run, it inserts some mock data into the DB. Then, at the end, it removes that mock data.

This PR will probably get merged before your test PR, so we can `rebase` your test PR branch so you can get these new tests.

test/backend_tests/test_resolve_fetch_failures.py
```diff
@@ -0,0 +1,32 @@
...
+PROJECT_ROOT = Path(TEST_DIR).parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+
+class TestBackendResolveFetchFailures(FetchAuditTestRunner):
+    """Inherits from the FetchAudit test class to get setup/teardown methods."""
+    # TODO: Need a better test. This one actually passes even though lots of members. i need one that has more than 10k
+    #  items
+    # Hope Termhub Test (v3) https://unite.nih.gov/workspace/module/view/latest/ri.workshop.main.module.5a6c64c0-e82b-4cf8-ba5b-645cd77a1dbf
+    mock_data = [{'table': 'code_sets', 'primary_key': 946809371, 'status_initially': 'fail-excessive-items',
```
joeflack4, line 22: @Sigfried Just wanna show you something strange. I was trying to come up with a case where something was actually failing. I found this case:

LongCovid (v1) (`112182256`)
```
{
   "properties": {
     "updateMessage": "Long Covid features",
     "lastModifiedAt": "2023-06-27T01:24:03.447Z",
     "sourceApplicationVersion": "2.0",
     "lastModifiedBy": "3b4fee14-4577-4fb3-98aa-f25e14fb0a18",
     "createdBy": "3b4fee14-4577-4fb3-98aa-f25e14fb0a18",
     "version": 1,
     "createdAt": "2023-06-27T00:11:50.735Z",
     "project": "RP-92E57A",
     "sourceApplication": "UNITE",
     "provenance": "Long Covid features",
     "conceptSetNameOMOP": "LongCovid",
     "limitations": "included from paper",
     "intention": "Narrow (specific)",
     "conceptSetVersionTitle": "LongCovid (v1)",
     "codesetId": 112182256,
     "isDraft": false,
     "isMostRecentVersion": true
   },
   "rid": 
 "ri.phonograph2-objects.main.object.241458e2-d61f-4f9a-9409-b7e0b6b96a68"
 }
```

However, when I look it up in the enclave, [nothing comes up](https://unite.nih.gov/workspace/module/view/latest/ri.workshop.main.module.5a6c64c0-e82b-4cf8-ba5b-645cd77a1dbf):
> No Concept Sets found matching the Search and Filtering Criteria

And when I try to fetch its container to see if it's archived, I don't see a container either. This returns an empty list:
```py
    make_objects_request('OMOPConceptSetContainer', query_params={'properties.conceptSetId.eq': _id}, verbose=verbose, return_type='data')
```

Just curious if you have any idea why that might be?

joeflack4: @Sigfried I ran a fetch since January 1st and so far this one above is actually the only failure case. Going to run again to be sure.

enclave_wrangler/objects_api.py
```diff
@@ -351,60 +357,44 @@ def fetch_cset_and_member_objects(
...
+            fetch_statuses.append({'table': 'code_sets', 'primary_key': version, 'status_initially':
+                'fail-excessive-members', 'comment': f"Failed after {len(cset['member_items'])} members."})
+        except Exception as err:
+            insert_fetch_status(fetch_statuses)
+            raise err
+        # todo: consider adding `and cset['expression_items']:` to further avoid false positives, or is the isDraft
+        #  check sufficient? I think it's correct that items are available immediately upon creation, but members are
+        #  not? Can a draft be finalized and have 0 expression items? Another thing to check would be createdAt >1+hours
+        #  as proxy; e.g. 'createdAt': '2023-02-14T15:10:13.921Z',
+        if not cset['member_items'] and not cset['properties']['isDraft']:
```
joeflack4, line 388: @Sigfried Regarding the issue where a cset is new and thus has 0 members when we fetch it:

I found an edge cases where there are 0 members, but it's OK and we don't need to flag:
1.  The cset has no members, but is a draft

Also, there are 2 other non-problematic cases I thought of that I could also check for in this if statement:
2. The cset has no members, but also has no expression items (_@Sigfried I think this can't happen, so I didn't add logic to check for it, but I will add if it can. Can it?)_
3. The cset has no members, but was created 'x' hours ago (_@Sigfried this is another way we can check things, but I thought it's better not to add this, as '1' and '2' should be sufficient to catch cases._)

backend/utils.py
```diff
@@ -32,12 +37,37 @@ def commify(n):
     return f'{n:,}'
 
 
+def call_github_action(event_type: str) -> Response:
```
joeflack4, line 40: @Sigfried Created this new function. Now used in two places: (i) the current refresh button implementation, (ii) during the db refresh, if it encounters data it can't fetch, it will trigger the solution in this PR: `resolve_fetch_failures.py`

Called like: `call_github_action('resolve-fetch-failures')`

Because of the change with 'i', I want to double check that this refactor didn't break the button. I don't have time for that today, but I've added it to my list of todo's.

### Coordinated deployments & testing
Though I also wonder if we should, as a team, do coordinated deployments. Like we can make an issue and name/tag it with the TermHub version we're deploying. Then list out all of the things that we want to test and make sure work when we're deploying to dev. Then only after all those things work do we deploy to prod.

Issue #341, opened 2023-05-26
DB Refresh: Via objects API: Non cset objects #105


joeflack4:
## Overview
This involves refreshing tables that are not populated by one of the core cset objects, e.g. (`OMOPConcept`, `OMOPConceptSet`, `OMOPConceptSetContainer`, `OmopConceptSetVersionItem`).

This includes
- `researcher`
- `project`

Sigfried:
@joeflack4, is this still not done? It should be a very quick fix. For researcher, it will be a problem if it's not updated, but may as well handle project at the same time.

joeflack4:
@Sigfried Still not done. It should not be a time consuming one, but I have it in line [after fixing vocab/counts & metadata updates](https://github.com/jhu-bids/TermHub/issues/398). Feel free to work on this as well but we should probably both be focused on testing.

PR #469, opened 2023-07-05
Pull request: simultaneous db requests, calling backend directly

### LINKED ISSUE #395: DB refresh: Simultaneous requests
## Overview
When a refresh request is made, should update `last_refresh_request` with the datetime that the request was made. When a refresh is complete, it currently updates the `last_refreshed_DB` with the datetime the refresh script started running / when it fetched data, (not when the script finished). 

If a request is made and `refresh_status` is `inactive`, we should start the refresh. If it is `active`, we should not start the refresh and exit out.

When the refresh script finishes, it should check `last_refresh_request`. If it has changed / is newer than when the script started, it should immediately begin another refresh, and repeat this indefinitely if this continues to be the case.

## Depends on
- #400

hugwuoke:
### Summary 

1. An if statement to decide whether or not to run the program
2. If the program doesn't run variable is saved that instructs the program to run again
3. At the end of the program it checks the variable, deletes and runs the function again. 

So refresh() becomes recursive if the function is called while already running.

### DIFF
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -35,18 +35,22 @@ def refresh_db(
     when the last refresh occurred, this will result in there being a gap in which any changes that occurred during that
     time will not be fetched, resulting in an incomplete database. Therefore by default this will raise an error unless
     this is set to True.
-    todo: Can update update_db_status_var() so that it can accept optional param 'con' to improve performance."""
+    todo: Can update update_db_status_var() so that it can accept optional param 'con' to improve performance.
+    todo: refactor `new_request_while_refreshing` usage for brevity in code and DB: Rather than checking a variable
+     new_request_while_refreshing , at the end of the refresh, if the last_refresh_request has changed / is newer than
+     what was set at the beginning of the script, it knows that a new request while refreshing has occurred, and use
+     that information to know to start new refresh. https://github.com/jhu-bids/TermHub/pull/469#discussion_r1253581138
+    """
     local = use_local_db
     print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
     t0, t0_str = datetime.now(), current_datetime()
-
     if check_db_status_var('refresh_status') == 'active':
+        print('INFO: Refresh already in progress. When that process completes, it will restart again. Exiting.')
         update_db_status_var('new_request_while_refreshing', t0_str, local)
         return
-
-    print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
     update_db_status_var('refresh_status', 'active', local)
     update_db_status_var('last_refresh_request', current_datetime(), local)
+
     try:
         with get_db_connection(local=local) as con:
             last_refresh = last_refresh_timestamp(con)
@@ -54,17 +58,18 @@ def refresh_db(
                 raise ValueError(SINCE_ERR)
             since = since if since else last_refresh
 
-            # Refresh DB:
-            # todo: will use instead when ready: all_new_objects_enclave_to_db()
-            if not csets_and_members_enclave_to_db(con, schema, since):
-                print('INFO: Database is up to date, no refresh necessary')
-                return
+            # Refresh DB
+            # todo: when ready, will use all_new_objects_enclave_to_db() instead of csets_and_members_enclave_to_db()
+            new_data: bool = csets_and_members_enclave_to_db(con, schema, since)
 
-        counts_update('DB refresh.', schema, local)
         update_db_status_var('refresh_status', 'inactive', local)
         update_db_status_var('last_refresh_success', current_datetime(), local)
         update_db_status_var('last_refresh_result', 'success', local)
-        print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
+        if new_data:
+            counts_update('DB refresh.', schema, local)
+            print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
+        else:
+            print('INFO: Database is up to date, no refresh necessary')
 
     except Exception as err:
         update_db_status_var('last_refresh_result', 'error', local)
@@ -74,9 +79,9 @@ def refresh_db(
         raise err
 
     if check_db_status_var('new_request_while_refreshing'):
-        # possibly look at the timestamp for sanity check
+        print('INFO: New refresh request detected while refresh was running. Starting a new refresh.')
         delete_db_status_var('new_request_while_refreshing')
-        refresh_db(use_local_database=use_local_database)
+        refresh_db(None, use_local_db, schema, force_non_contiguity)
 
 
 def cli():
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -40,41 +40,38 @@ def refresh_db(
     print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
     t0, t0_str = datetime.now(), current_datetime()
 
-    with get_db_connection(local=local) as con:
-        last_refresh = last_refresh_timestamp(con)
-        if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
-            raise ValueError(SINCE_ERR)
-        since = since if since else last_refresh
+    if check_db_status_var('refresh_status') == 'active':
+        update_db_status_var('new_request_while_refreshing', t0_str, local)
+        return
+    
+    print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
+    update_db_status_var('refresh_status', 'active', local)
+    update_db_status_var('last_refresh_request', current_datetime(), local)
+    try:
+        with get_db_connection(local=local) as con:
+            last_refresh = last_refresh_timestamp(con)
+            if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
+                raise ValueError(SINCE_ERR)
+            since = since if since else last_refresh
 
-        #todo: new way of checking db is active
-        '''if not is_new_db_data(since):
-            'INFO: Database is up to date, no refresh necessary'
-            return'''
-
-        if check_db_status_var('refresh_status') == 'active':
-            update_db_status_var('new_request_while_refreshing', t0_str, local)
-            return
-
-        print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
-        update_db_status_var('refresh_status', 'active', local)
-        update_db_status_var('last_refresh_request', current_datetime(), local)
-        try:
             # Refresh DB:
             # todo: will use instead when ready: all_new_objects_enclave_to_db()
-            csets_and_members_enclave_to_db(con, schema, since)
+            if not csets_and_members_enclave_to_db(con, schema, since):
+                print('INFO: Database is up to date, no refresh necessary')
+                return
 
-            counts_update('DB refresh.', schema, local)
-            update_db_status_var('refresh_status', 'inactive', local)
-            update_db_status_var('last_refresh_success', current_datetime(), local)
-            update_db_status_var('last_refresh_result', 'success', local)
-            print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
+        counts_update('DB refresh.', schema, local)
+        update_db_status_var('refresh_status', 'inactive', local)
+        update_db_status_var('last_refresh_success', current_datetime(), local)
+        update_db_status_var('last_refresh_result', 'success', local)
+        print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
 
-        except Exception as err:
-            update_db_status_var('last_refresh_result', 'error', local)
-            update_db_status_var('refresh_status', 'inactive', local)
-            counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
-            print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
-            raise err
+    except Exception as err:
+        update_db_status_var('last_refresh_result', 'error', local)
+        update_db_status_var('refresh_status', 'inactive', local)
+        counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
+        print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
+        raise err
 
     if check_db_status_var('new_request_while_refreshing'):
         # possibly look at the timestamp for sanity check
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -14,7 +14,7 @@ sys.path.insert(0, str(PROJECT_ROOT))
 from backend.db.analysis import counts_update,counts_docs
 from backend.db.config import CONFIG
 from backend.db.utils import current_datetime, get_db_connection, last_refresh_timestamp, update_db_status_var,check_db_status_var, delete_db_status_var
-from enclave_wrangler.objects_api import csets_and_members_enclave_to_db
+from enclave_wrangler.objects_api import csets_and_members_enclave_to_db,is_new_db_data
 
 DESC = 'Refresh TermHub database w/ newest updates from the Enclave using the objects API.'
 SINCE_ERR = '--since is more recent than the database\'s record of last refresh, which will result in missing data.'
@@ -40,42 +40,46 @@ def refresh_db(
     print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
     t0, t0_str = datetime.now(), current_datetime()
 
-    if check_db_status_var('refresh_status') == 'active':
-        update_db_status_var('new_request_while_refreshing', t0_str, local)
-        return
+    with get_db_connection(local=local) as con:
+        last_refresh = last_refresh_timestamp(con)
+        if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
+            raise ValueError(SINCE_ERR)
+        since = since if since else last_refresh
 
-    update_db_status_var('refresh_status', 'active', local)
-    update_db_status_var('last_refresh_request', current_datetime(), local)
-    try:
-        with get_db_connection(local=local) as con:
-            last_refresh = last_refresh_timestamp(con)
-            if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
-                raise ValueError(SINCE_ERR)
-            since = since if since else last_refresh
+        #todo: new way of checking db is active
+        '''if not is_new_db_data(since):
+            'INFO: Database is up to date, no refresh necessary'
+            return'''
 
-            # Refresh DB
+        if check_db_status_var('refresh_status') == 'active':
+            update_db_status_var('new_request_while_refreshing', t0_str, local)
+            return
+
+        print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
+        update_db_status_var('refresh_status', 'active', local)
+        update_db_status_var('last_refresh_request', current_datetime(), local)
+        try:
+            # Refresh DB:
             # todo: will use instead when ready: all_new_objects_enclave_to_db()
             csets_and_members_enclave_to_db(con, schema, since)
 
-        counts_update('DB refresh.', schema, local)
-        update_db_status_var('refresh_status', 'inactive', local)
-        update_db_status_var('last_refresh_success', current_datetime(), local)
-        update_db_status_var('last_refresh_result', 'success', local)
-        counts_docs()
-        print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
+            counts_update('DB refresh.', schema, local)
+            update_db_status_var('refresh_status', 'inactive', local)
+            update_db_status_var('last_refresh_success', current_datetime(), local)
+            update_db_status_var('last_refresh_result', 'success', local)
+            print(f'INFO: Database refresh complete in {(datetime.now() - t0).seconds} seconds.')
 
-    except Exception as err:
-        update_db_status_var('last_refresh_result', 'error', local)
-        update_db_status_var('refresh_status', 'inactive', local)
-        counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
-        print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
-        raise err
+        except Exception as err:
+            update_db_status_var('last_refresh_result', 'error', local)
+            update_db_status_var('refresh_status', 'inactive', local)
+            counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
+            print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
+            raise err
 
     if check_db_status_var('new_request_while_refreshing'):
         # possibly look at the timestamp for sanity check
         delete_db_status_var('new_request_while_refreshing')
         refresh_db(use_local_database=use_local_database)
-
 
 
 def cli():
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -13,8 +13,8 @@ PROJECT_ROOT = os.path.join(BACKEND_DIR, '..')
 sys.path.insert(0, str(PROJECT_ROOT))
 from backend.db.analysis import counts_update,counts_docs
 from backend.db.config import CONFIG
-from backend.db.utils import current_datetime, get_db_connection, last_refresh_timestamp, update_db_status_var,check_db_status_var, delete_db_status_var
-from enclave_wrangler.objects_api import csets_and_members_enclave_to_db,is_new_db_data
+from backend.db.utils import current_datetime, get_db_connection, last_refresh_timestamp, update_db_status_var, check_db_status_var, delete_db_status_var
+from enclave_wrangler.objects_api import csets_and_members_enclave_to_db
 
 DESC = 'Refresh TermHub database w/ newest updates from the Enclave using the objects API.'
 SINCE_ERR = '--since is more recent than the database\'s record of last refresh, which will result in missing data.'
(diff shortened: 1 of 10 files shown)


backend/db/utils.py
```diff
@@ -156,14 +156,24 @@ def is_table_up_to_date(table_name: str, skip_if_updated_within_hours: int = Non
     return check_if_updated(last_updated_key, skip_if_updated_within_hours)
 
 
-def update_db_status_var(key: str, val: str, local=False):
+def delete_db_status_var(key: str, local=False):
```
joeflack4, line 159: Thanks for adding this. Just a couple cleanup steps:
- Move this todo `todo: change to a 1-liner UPDATE statement` from `delete_db_status_var()` into the `update_db_status_var()` docstring
- Update the `delete_db_status_var()` docstring to something like "Delete a variable from the `manage` table.

backend/db/utils.py
```diff
@@ -156,14 +156,24 @@ def is_table_up_to_date(table_name: str, skip_if_updated_within_hours: int = Non
...
+    """Update the `manage` table with information for a given variable, e.g. when a table was last updated"""
+    with get_db_connection(schema='', local=local) as con2:
+        run_sql(con2, f"DELETE FROM public.manage WHERE key = '{key}';")
         sql_str = f"INSERT INTO public.manage (key, value) VALUES (:key, :val);"
         run_sql(con2, sql_str, {'key': key, 'val': val})
 
+def check_db_status_var(key: str,  local=False):
+    """Check the value of a given variable the `manage`table """
+    with get_db_connection(schema='', local=local) as con2:
+        return sql_query_single_col(con2, f"SELECT * FROM public.manage WHERE key = '{key}';")
```
joeflack4, line 176: Thanks for adding this! One change to make is to change select from `SELECT *` to `SELECT value` (as in `last_refresh_timestamp()`).

enclave_wrangler/utils.py
```diff
@@ -157,30 +157,38 @@ def response_failed(response: Response) -> bool:
...
 
 def handle_response_error(
-    response: Response, error_dir: Union[str, None] = None, fail: bool = False, calling_func: str = 'response error'
+    response: Response, error_dir: Union[str, None] = None, calling_func: str = 'response error'
 ):
     """This code was taken out of make_objects_request() (formerly get_objects_by_type()). Trying to use it
-    for all Response errors now. Not sure if it's entirely appropriate."""
+    for all Response errors now. Not sure if it's entirely appropriate.
+
+    TODO: Test this! Siggie and Hope made changes in late June and we didn't test them. If this breaks, blame them
```
joeflack4, line 165: Just commenting that I've seen these changes to error reporting! Fine with me. Hope it works. If you'd like, you can prioritize adding a test for this ahead of #464. I leave it up to you.

backend/db/refresh.py
```diff
@@ -64,6 +69,13 @@ def refresh_db(
         print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
         raise err
 
+    if check_db_status_var('new_request_while_refreshing'):
+        # possibly look at the timestamp for sanity check
+        delete_db_status_var('new_request_while_refreshing')
+        refresh_db(use_local_database=use_local_database)
```
joeflack4, line 75: @hugwuoke I think the way you've set this up will work. Thanks much!
It looks like this is ready for me to take it on a test run now, as the PR is not a draft. Is that correct? If so I'll make sure it works and merge~

backend/db/refresh.py
```diff
@@ -38,8 +38,13 @@ def refresh_db(
     local = use_local_database
     print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
     t0, t0_str = datetime.now(), current_datetime()
-    update_db_status_var('last_refresh_request', t0_str, local)
+
+    if check_db_status_var('refresh_status') == 'active':
+        update_db_status_var('new_request_while_refreshing', t0_str, local)
```
joeflack4, line 43: I think the way you've done this will work. My initial thought of doing this was:
- If refresh status is active, exit out (as you've done)
- Rather than checking a variable `new_request_while_refreshing `, at the end of the refresh, if the `last_refresh_request` has changed / is newer than what was set at the beginning of the script, it knows that a new request while refreshing has occurred, and use that information to know to start a new refresh.

Depending on time, while I'm testing this out I may refactor this to what I've described above. But I'm also happy with what you have!

backend/db/refresh.py
hugwuoke: Yes I would say its ready !

backend/db/refresh.py
```diff
@@ -36,33 +36,48 @@ def refresh_db(
...
-            if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
-                raise ValueError(SINCE_ERR)
-            since = since if since else last_refresh
 
+    with get_db_connection(local=local) as con:
+        last_refresh = last_refresh_timestamp(con)
+        if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
+            raise ValueError(SINCE_ERR)
+        since = since if since else last_refresh
+
```
hugwuoke, line 46: I moved this entire part of the code up, because when working with Siggie it seemed like making as few connections as possible in a function was a priority for them. On running in debugger it seemed like this change didn't affect the program's functionality.

enclave_wrangler/objects_api.py
```diff
@@ -405,7 +405,15 @@ def fetch_cset_and_member_objects(
 
     return {'OMOPConceptSetContainer': cset_containers, 'OMOPConceptSet': cset_versions,
             'OmopConceptSetVersionItem': flat_expression_items, 'OMOPConcept': flat_member_items}
-
+def is_new_db_data(since: Union[datetime, str]):
+    "Returns True if there is new data for the db, false if not"
```
hugwuoke, line 409: I created a new function to test for new data, I'm not sure this file is the best place for it.

joeflack4: It's OK to be there right now. I'm going to be moving a lot of stuff from inside this file into `backend/db/` instead, at some point soon probably. I can move this at that time.

I'm supposing that this function will call the first part of `fetch_cset_and_member_objects()`. This function returns this structure of new stuff to put in our DB (copy/pasting from its docstring):

```
    :return
      - cset containers
      - cset versions
          - member items
          - expression items
      - member items
      - expression items
```

Right now the way it's set up, it fetches cset versions and, based on that, it fetches all of these other things as they relate to these cset versions. Siggie wants me to change it to fetch containers independently of cset versions, so I'll be changing that soon.

Anyway, let's say that it tries to fetch new cset versions since `since`, but finds nothing. I believe the way that it will turn out is that this:
```py
    return {'OMOPConceptSetContainer': cset_containers, 'OMOPConceptSet': cset_versions,
            'OmopConceptSetVersionItem': flat_expression_items, 'OMOPConcept': flat_member_items}
```

will resolve to this:
```py
    return {'OMOPConceptSetContainer': [], 'OMOPConceptSet': [],
            'OmopConceptSetVersionItem': [], 'OMOPConcept': []}
```

Honestly though, if there's no new data, I should just have it return `None`.

In any case, I guess what I'm thinking is that you can have `is_new_db_data()` simply call `fetch_cset_and_member_objects()` and return true or false based on what it sees in response. But another alternative is that we can just skip `is_new_db_data()` and use `fetch_cset_and_member_objects()` in its place. It kinda depends on why this function is being created / what it will be used for. It wouldn't be performant to fetch a bunch of stuff, just to see if there is any data in the DB, just fo tech it again.

Although, another thing you could do is to tell if there's new data is to just call a subset of `fetch_cset_and_member_objects()`. As I was mentioning earlier, it checks for cset versions and then fetches other stuff based on that. And soon will be changed to also fetch containers whether or not there are any new cset versions. So your function could just fetch those two things independently and return true/false depending on if anything comes back:

```py
cset_versions: List[Dict] = fetch_objects_since_datetime('OMOPConceptSet', since)
cset_containers: List[Dict] = fetch_objects_since_datetime('OMOPConceptSetContainer', since)
return True if cset_versions or cset_containers else False
```

Then again, maybe there is some other way to check if there's new data for us in the enclave other than what I've set up now, like a specific endpoint in the enclave API, but I haven't come across such alternative AFAIK.

backend/db/refresh.py
```diff
@@ -36,46 +36,49 @@ def refresh_db(
...
-    update_db_status_var('refresh_status', 'active', local)
-    update_db_status_var('last_refresh_request', current_datetime(), local)
-    try:
-        with get_db_connection(local=local) as con:
-            last_refresh = last_refresh_timestamp(con)
-            if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
-                raise ValueError(SINCE_ERR)
-            since = since if since else last_refresh
+        if not is_new_db_data(since):
+            'INFO: Database is up to date, no refresh necessary'
```
joeflack4, line 48: I like this change! Kinda silly I didn't add it before.

I mentioned some ideas in [this other comment](https://github.com/jhu-bids/TermHub/pull/469#discussion_r1256447069).

I didn't label these options in my comment, but I'll call this option "A":
You could simply determine if there's any data during the normal call to `fetch_cset_and_member_objects()`.

```py
def csets_and_members_enclave_to_db(
    con: Connection, schema: str, since: Union[datetime, str], filter_0_member_sets=True
):
    """Fetch new csets and members, if needed, and then update database with them."""
    print('Fetching new data from the N3C data enclave...')
    t0 = datetime.now()
    csets_and_members: Dict[str, List[Dict]] = fetch_cset_and_member_objects(since)
```

After that last line, you could check if there's any new data, and if not, simply return `None`. This will return execution back to `refresh_db()` in `refresh.py`, where you could check if `None` is returned, and then exit out.

Option B:
I proposed an implementation of `is_new_db_data()`. You could use that and keep where you're currently calling `is_new_db_data()` in `refresh.py`.

Or Option C:
Maybe you'll think of something I haven't thought of!

backend/db/refresh.py
joeflack4: Probably not a bad idea.

backend/routes/db.py
```diff
@@ -356,19 +357,7 @@ def _cset_members_items(codeset_ids: Union[str, None] = Query(default=''), ) ->
...
-        "Content-Type": "application/json"
-    }
-    # args = {"event-type":"refresh-db"}
-    url = 'https://api.github.com/repos/jhu-bids/TermHub/dispatches'
-    payload = {
-        'event_type': 'refresh-db',
-    }
-    response = requests.post(url, headers = headers,data=json.dumps(payload))
-    return response
+    os.system('python backend/db/refresh.py')
```
hugwuoke, line 361: @joeflack4 I added this to call refresh directly. I want to go with your suggestion of having fetch_cset_and_members_objects returning none as a way to check that there is no new data. After I change that I want to test this again to make sure it's working.

joeflack4: Sounds good.

As for running the refresh script, what I would do here is import `refresh_db` and run that function directly. That should have the same effect as running the script as a process the way you're doing it, with the added benefits of (i) we can debug it if we need to, (ii) if there's a failure, we'll see the stacktrace, etc.

Also and FYI, concerning subprocesses, I haven't seen `os.system` before. I just tried it and I do see that it runs a system call. I've always used `subprocess.Popen` or `subprocess.call` (I prefer the latter). IDK about `os.system`, but you can get the results of these by like `results = subprocess.call()` which will allow you to see anything that was printed to the console while the process was running. It's a little wonkier than calling a function directly, but sometimes it's the only option.

backend/db/refresh.py
```diff
@@ -39,18 +39,26 @@ def refresh_db(
...
                 raise ValueError(SINCE_ERR)
             since = since if since else last_refresh
 
-            # Refresh DB
+            # Refresh DB:
             # todo: will use instead when ready: all_new_objects_enclave_to_db()
-            csets_and_members_enclave_to_db(con, schema, since)
+            if not csets_and_members_enclave_to_db(con, schema, since):
+                print('INFO: Database is up to date, no refresh necessary')
+                return
```
hugwuoke, line 62: This was my idea for exiting out, I changed this function to return None when no refresh is necessary as suggested so it exits out. It should run as normal and update as expected if none is not returned.

joeflack4: This change LGTM!

backend/db/utils.py
joeflack4: Hope probably made this change already. I should double check.

enclave_wrangler/objects_api.py
joeflack4: New function was removed based on other stuff I said.

hugwuoke: I actually changed it to return none in fetch_cset_and_members based on the the two lines of code you wrote, then I changed all functions that call fetch_cset_and_members to be able to handle that. 

```python
cset_versions: List[Dict] = fetch_objects_since_datetime('OMOPConceptSet', since)
cset_containers: List[Dict] = fetch_objects_since_datetime('OMOPConceptSetContainer', since) ```

File: backend/db/utils.py (hunk 1 above)
Comment (joeflack4) [reply backend/db/utils.py]: Confirmed these changes are made, thanks Hope!

File: backend/db/utils.py (hunk 2 above)
Comment (joeflack4) [reply backend/db/utils.py]: Confirming change made! Thank you!

File: enclave_wrangler/objects_api.py (hunk 7 above)
Comment (joeflack4) [reply enclave_wrangler/objects_api.py]: Confirming that `None` is being returned in the correct way in `fetch_cset_and_member_objects()`, thanks!
```py
    if not cset_containers and not cset_versions:
        return None
```

You also correctly made this change to `csets_and_members_enclave_to_db()`:
```py
   if not csets_and_members:
        return None
```

This one is a mistake, though!
```py
def fetch_all_new_objects():
    csets_and_members: Dict[str, List] = fetch_cset_and_member_objects(since)
    if not csets_and_members:
        return None
    researchers = get_researchers()
    projects = get_projects()
```

I think it's a good thing that you looked at where `fetch_cset_and_member_objects()` was being called, and made these changes. 
And it is not a big deal that this is a mistake, because we're not even using this function `fetch_all_new_objects()` yet. Right now we're going from `refresh_db()` -> `csets_and_members_enclave_to_db()` -> `fetch_cset_and_member_objects()`
But eventually when we get to #341 we will go from `refresh_db()` -> `all_new_objects_enclave_to_db()` -> `fetch_all_new_objects()`, which will also call -> `fetch_cset_and_member_objects()`
And the reason it's not correct in this function to return if no `csets_and_members`, is just because there are no new concept sets, etc, from that function, that does not mean that there is not new researcher and project data.
I went ahead and fixed this and pushed so you don't need to worry about on your end!

backend/db/refresh.py
```diff
@@ -65,6 +73,12 @@ def refresh_db(
         print(f"Database refresh incomplete. An exception occurred.", file=sys.stderr)
         raise err
 
+    if check_db_status_var('new_request_while_refreshing'):
+        # possibly look at the timestamp for sanity check
+        delete_db_status_var('new_request_while_refreshing')
+        refresh_db(use_local_database=use_local_database)
```
joeflack4, line 79: Noticed a bug here! Should be: `use_local_db`. Fixing now!
Screen Shot 2023-07-20 at 6 12 34 PM

I think I'm going to largely replace `use_local_database` with `use_local_db` just because it is shorter.

backend/db/refresh.py
```diff
@@ -35,22 +35,35 @@ def refresh_db(
...
             if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
                 raise ValueError(SINCE_ERR)
             since = since if since else last_refresh
 
-            # Refresh DB
-            # todo: will use instead when ready: all_new_objects_enclave_to_db()
-            csets_and_members_enclave_to_db(con, schema, since)
+            # Refresh DB:
+            # todo: when ready, will use all_new_objects_enclave_to_db() instead of csets_and_members_enclave_to_db()
+            if not csets_and_members_enclave_to_db(con, schema, since):
```
joeflack4, line 64: ### Closing out refresh if no new updates
I just noticed a few things that needed to be changed here!

1. Bugfix: `csets_and_members_enclave_to_db()` was actually always returning `None`!
The last line of the function was `return csets_and_members_to_db(con, schema, csets_and_members)`
But that function actually had no `return` statement. Basically when a function has no `return`, it's the same as that function returning `None`. So it was always returning `None`! I changed it to return `'success'` at the very end.

2. Updating some status variables
The way you had it, if `csets_and_members_enclave_to_db()` returned `None`, and then `refresh_db()` would exit out. But, this would result in a few of the DB variables not being updated. For example, `refresh_status` would still erroneously be in the state `active`. And I also felt it would be good to update `last_refresh_success` and `last_refresh_result`. So really, the only thing it doesn't do now when there were no new updates is that it doesn't call `counts_update()`.

3. Allowing it to do  `check_db_status_var('new_request_while_refreshing')`
So the way you had it, if nothing was returned `csets_and_members_enclave_to_db()`, it was exiting before it could get here. Realistically though, if there were no new database updates, and if the user clicked the button / there was another refresh request during that time, it is very likely that even if we check again, there will be no updates. Still though, it is possible that there could be some updates, and we also want to close out and at least run it and do `delete_db_status_var('new_request_while_refreshing')`,.

enclave_wrangler/objects_api.py
```diff
@@ -288,21 +288,23 @@ def all_new_objects_enclave_to_db(since: Union[datetime, str]) -> Dict[str, List
...
-    con: Connection, schema: str, since: Union[datetime, str], filter_0_member_sets=True
-):
-    """Fetch new csets and members, if needed, and then update database with them."""
+def csets_and_members_enclave_to_db(con: Connection, schema: str, since: Union[datetime, str]) -> Union[None, str]:
+    """Fetch new csets and members, if needed, and then update database with them, returns None is no updates needed"""
     print('Fetching new data from the N3C data enclave...')
     t0 = datetime.now()
     csets_and_members: Dict[str, List[Dict]] = fetch_cset_and_member_objects(since)
+    if not csets_and_members:
+        return None
```
joeflack4, line 297: I actually realized that `return None` is the same as `return`. They both return `None`. So I shortened this to just `return`.

enclave_wrangler/objects_api.py
```diff
@@ -354,6 +355,10 @@ def fetch_cset_and_member_objects(
                              f'for that specific container.')
         cset_containers.append(container[0]['properties'])
 
+    # Returning None if no data
+    if not cset_containers and not cset_versions:
+        return None
```
joeflack4, line 360: I actually realized that `return None` is the same as `return`. They both return `None`. So I shortened this to just `return`.

backend/db/utils.py
```diff
@@ -157,16 +157,22 @@ def is_table_up_to_date(table_name: str, skip_if_updated_within_hours: int = Non
...
     return check_if_updated(last_updated_key, skip_if_updated_within_hours)
 
-
 # todo: Can update update_db_status_var() so that it can accept optional param 'con' to improve performance.
 def update_db_status_var(key: str, val: str, local=False):
     """Update the `manage` table with information for a given variable, e.g. when a table was last updated
     todo: change to a 1-liner UPDATE statement"""
     with get_db_connection(schema='', local=local) as con:
         run_sql(con, f"DELETE FROM public.manage WHERE key = '{key}';")
-        sql_str = f"INSERT INTO public.manage (key, value) VALUES (:key, :val);"
```
joeflack4, line 167: @hugwuoke Uhoh. I think when you were pair programming with Siggie you guys might have seen that I had a todo here "change to a 1-liner UPDATE statement". It looks like you guys were trying to do that, and ended up deleting the line where it inserts a new value for the db var, but only leaving the 1 line where it deletes the variable. When that change was made, I think it was meant to be changed from a `DELETE` to an `UPDATE`.

It probably doesn't take very long to change it fo an `UPDATE`, but I'm a little strapped for time here today as well, so I'm just going to revert it back to the way it was before: delete the old variable, insert the variable again w/ its new value.

backend/db/utils.py
```diff
@@ -167,6 +166,16 @@ def update_db_status_var(key: str, val: str, local=False):
         sql_str = f"INSERT INTO public.manage (key, value) VALUES (:key, :val);"
         run_sql(con, sql_str, {'key': key, 'val': val})
 
+def check_db_status_var(key: str,  local=False):
+    """Check the value of a given variable the `manage`table """
+    with get_db_connection(schema='', local=local) as con2:
+        results: List = sql_query_single_col(con2, f"SELECT value FROM public.manage WHERE key = '{key}';")
+        return results[0] if results else None
```
joeflack4, line 173: @hugwuoke This one would have been impossible to catch without doing a test run, which is my job for this PR. I think there were some recent refactoring changes. I think `sql_query_single_col()` is a new column, and it returns a list, so I had to grab the 0th item.

backend/routes/db.py
```diff
@@ -343,12 +342,19 @@ def _cset_members_items(codeset_ids: Union[str, None] = Query(default=''), ) ->
     codeset_ids: List[int] = parse_codeset_ids(codeset_ids)
     return get_cset_members_items(codeset_ids)
 
-
 @router.get("/db-refresh")
-def db_refresh_route() -> Response:
-    """Triggers refresh of the database"""
-    response: Response = call_github_action('refresh-db')
-    return response
+def db_refresh_route():
```
joeflack4, line 346: ### Change back to GitHub action?
@Sigfried I thought this change to call backend directly rather than GH action. I'm going to merge this now, as we've decided. But I think maybe we should go back to the GH action for the following reasons:

1. Easier to check logs
If there's a problem, I can go to the actions tab and find easily. Finding on azure takes many more clicks, and then I have to scroll up to find where the refresh got logged, and it may be mixed with logs for other requests.
2. Almost always won't increase speed of refresh
This was supposed to be the only benefit of calling it directly.
If someone creates a cset and clicks the button, it doesn't matter that our backend is faster than a GH action starting up, since it will take 20-45 minutes for that cset to be ready anyway. so effectively this is not faster, except for fetching csets that are not brand new, which (i) is not the primary thing people are using the button for, and (ii) is unlikely to happen w/ a fast refresh rate.
3. Harder to make changes
If we want to make changes to the refresh, we have to redeploy the whole app to make that work, rather than pushing to develop.
4. Uses more server resources.
5. Possible server stability issues
I'm not sure, but I wonder if there is some edge case where an error that happens during the refresh, or other unanticipated side effects, could have an effect on performance or stability of the web server.
This is on top of the

joeflack4 (commented):
@hugwuoke I'm merging this now! Thanks for all your work on this.
Can you also review my latest comments so you are aware of my thoughts and changes?
You can let me know you've reviewed them by responding if you have any questions or thoughts you'd like to share, or it's also fine to respond with an emote / thumbs up too.

@Sigfried @hugwuoke
I tested this feature locally. I never had an issue where I got an error when doing simultaneous requests. But the way they were being handled wasn't as expect either. Some of the times I tested it, it went as expected: the first refresh started, I hit button again to initiate refresh, and the secondary refresh exited out, while the first refresh saw the new request and started over again at the end. But some of the times I tested it, this didn't happen. Instead, the first refresh fully completed, and then the next refresh started and completed sequentially. I think this has to do with the nature of the local FastAPI server process not being properly multithreaded; not being quite the same as the server process that is running on dev/prod. So when we deploy to dev, we should test that this is working correctly.

In pursuit of this, I created a new milestone and issue for us to do these deployment checks:
- https://github.com/jhu-bids/TermHub/milestone/14
- #493

backend/db/utils.py
hugwuoke: I think some more clarification on how you tested might be helpful, I thought I tested things ok, I opened up the page locally and hit refresh and everything seemed fine. I think it might be helpful to get more information on how you tested so that I would be able to do more of that by myself in the future.

joeflack4: Hmm, you know what. It might be possible that it was working correctly when you were programming this. I did a `git rebase develop` on this branch, and I don't remember if I encountered this before the rebase, or afterwards.

To clarify, basically I needed to add `results[0]`. I believe this came up towards the beginning of the script think it was when it was checking if the status was `'active'`, but what was coming back from this function call was `['active']`.

So now that I think of it, this may well have come up before I rebased. But since it didn't produce an error message, that would be why it would have been difficult for you to catch.

I think I noticed this because I clicked the button once, and then I clicked it a second time. And I wanted to check that the second click was behaving correctly. Since it wasn't able to correctly determine that the refresh was in the `'active'` state, I noticed that the second click resulted in it not exiting out as expected (e.g. I didn't see the `print` statement about it exiting out). 

---

In any case, this was a bit tricky to test out. I am not 100% confident in my testing results. I tried this a few ways ways. The first way, I set a breakpoint in the middle of the refresh, clicked the refresh button, let it sit there paused at the breakpoint, and then I hit the refresh button again. Another way I tested it, was rather than setting a breakpoint, I did `time.sleep()` a bunch of times. And hit the refresh button a 2nd time while those sleeps were running.

FYI as a safety measure I also temporarily commented some lines of code so that it was not actually going to run the refresh for real and  I didn't want it to update some of the DB variables.

But yeah, I want to test this out when we deploy this to dev. I'm not 100% confident in my results because some of the times that I ran this, I got the result that I expected. I saw the 2nd refresh started and exited out as expected. But other times, 2nd refresh started sequentially only after the 1st had completed. It's a bit tricky to test this kind of thing because it relies on the multithreading nature of the FastAPI web server to be working correctly. And I believe that it functions differently in the development environment than when deployed.

backend/routes/db.py
Sigfried: How about this: when the api call is made, if backend is running locally, run directly; if on prod, run GH action; if on dev, flip a coin :)  

The button call's `db-refresh` route in any case. Just add some logic [here](https://github.com/jhu-bids/TermHub/blob/3ee2ee4b18a29724667c01986ebea2d565438e48/backend/routes/db.py#L359-L361)

joeflack4: I was thinking about that, except I think dev should call the GH action as well.

But for local, I don't know if it even makes sense. It will be the same as running the refresh script with no params, which you can do via right click + debug. So not the biggest priority for me.

Thanks for reading my points. I'll do this after some of the more critical db refresh priorities.

If for some reason members start coming through the enclave API faster, we can change it back as speed will matter more in that situtaion.

PR #486, opened 2023-07-14
Pull request: updating documentation with GitHub action. 

hugwuoke:
I did a google search and this is what I found, so I implemented it in the code but I didn't run it. 

https://github.com/marketplace/actions/git-auto-commit
https://github.com/stefanzweifel/git-auto-commit-action

### DIFF
--- a/.github/workflows/refresh_db.yml
+++ b/.github/workflows/refresh_db.yml
@@ -47,3 +47,8 @@ jobs:
       - name: Refresh db
         run: |
           python backend/db/refresh.py
+
+      - name: Commit changes to documentation
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+        commit_message: Updating documentation in docs/backend/db/analysis.md
--- a/docs/call_graph.md
+++ b/docs/call_graph.md
@@ -47,3 +47,4 @@ Most of the links don't work in pycharm but they work when used in GitHub.
       
 
 ### [dataset.py](../enclave_wrangler/datasets.py)
+


.github/workflows/refresh_db.yml
```diff
@@ -47,3 +47,8 @@ jobs:
       - name: Refresh db
         run: |
           python backend/db/refresh.py
+
+      - name: Commit changes to documentation
+      - uses: stefanzweifel/git-auto-commit-action@v4
```
joeflack4, line 52: @hugwuoke I must say, this is super thoughtful and proactive of you to take the initiative on this! 

I looked at the documentation, and does seem easy to use: https://github.com/stefanzweifel/git-auto-commit-action

I was thinking of doing a small test of this separate from what we really want to use it for (adding the `counts-docs` step to the end of the DB refresh). But that would be a bit messy I think. Since this involves a GitHub action, if we do this as a brand new action in a new PR, the action won't actually show up in https://github.com/jhu-bids/termhub/actions for us to run until it's in our main branch, `develop`.

On the other hand, since the DB refresh action already exists on `develop`, we can run it from the PR branch before merging:
Screen Shot 2023-07-14 at 5 25 14 PM

So let's do that; proceed with adding this to the DB refresh action and test it out while it's still a PR.

I'm thinking of the following steps:
1. **Remake this PR**: I see the changes to `supergroup` and `termhub-csets`. I don't know why these submodules get changes at random times, but I'm a bit leery of merging the changes without understanding why they're there. I clicked the files but it doesn't show me any diff. If we can remake it without these submodule changes, that would be good. You can either (a) create a new branch and re-open a new PR without the submodule changes (but this will delete our comment history), or (b) delete your local branch, then make a new branch w/ the same exact branch name `485-db-refresh-updating-documentation`, make your same changes, and then you can set the upstream branch and force push overwrite the current state of the branch via `git push -uf origin 485-db-refresh-updating-documentation`.
2. **Fix the failing action**: I see that this action was actually [run already and failed](https://github.com/jhu-bids/TermHub/actions/runs/5557925781). I am unsure how that happened. It says that it was triggered not by you manually running it, but by a git push. However, `refresh_db.yml` lacks `on: push` at the top of the file, so I don't see why it would have been triggered. Another strange thing is that the action name in the GitHub UI changed from "DB Refresh (Objects API)" to "[.github/workflows/refresh_db.yml](https://github.com/jhu-bids/TermHub/actions/workflows/refresh_db.yml)". I'm not too worried about that. I think it can be fixed, though I don't know why it's happening. In any case, about fixing the error. It says **every step must define a `uses` or `run` key**. I think what you need to do is replace `- uses: stefanzweifel/git-auto-commit-action@v4` with `  uses: stefanzweifel/git-auto-commit-action@v4`, effectively removing the `-` and leaving an indent.
3. **Call `counts-docs`**: Can be done one of two ways, though I am slightly preferring 'b': (a) You can add a separate step to the GitHub action that runs `make counts-docs`, or (b) in `refresh.py`, change `from backend.db.analysis import counts_update` to `from backend.db.analysis import counts_update, docs` and then run `docs()` (maybe I shoulda named this function `counts_docs()`; you can rename it if you want) in at the end of both the try and except blocks, right before the print statements.

Let me know what you think of this assessment / plan of action and feel free to proceed with it if you like!

hugwuoke: I made the changes and ran it and it worked !!

joeflack4: I am really glad to hear it worked I see [the commit](https://github.com/jhu-bids/TermHub/pull/486/commits/aedb83d8d207a0bdabdc2dbd7427bbd244d25cdd) too!. I just took a look at your code changes.

I want to merge this; it's probably alright, but I'm going to sit on it just a little bit more. One thing I noticed is regarding bullet (1) above; I still see `termhub-csets` and `supergroup` changes there. That one is the most tedious. Were you able to try one of my recommended methods; I suppose 'b'? If 'b' doesn't work we could try 'a'.

hugwuoke: I did do b, so I guess I'll go ahead and try a.

joeflack4: Alright! I'm sorry about the inconvenience. You can leave this PR open and we can close it later.

PR #497, opened 2023-07-27
DB Refresh: New csets with 0 members

joeflack4:
Addresses issue where we are trying to fetch a brand new concept set that has just been finalized within the last 45 minutes or so, but due to long processes that need to run in the Enclave, its concept set members were not available when we tried to fetch it, so this work will allow for a process that will continuously check until those members appear, and then we will integrate it into the DB.

Updates

    DB Refresh: New csets with 0 members
    Addresses issue where we are trying to fetch a brand new concept set that has just been finalized within the last 45 minutes or so, but due to long processes that need to run in the Enclave, its concept set members were not available when we tried to fetch it, so this work will allow for a process that will continuously check until those members appear, and then we will integrate it into the DB.
    - Add: New script for this
    - Add: New GH action
    - Update: Several functions updated to make this work
    - Update: Frontend: About page: Verbiage about the DB refresh

    DB Refresh: general
    - Update: fetch_audit table: Added primary keys
    - Bugfix: insert_fetch_status(): Was inserting duplicate rows because of a failed check if the record already existed. Because of generality of the fetch_audit table, int codeset_id needed to be converted to str before this check.
    - Add: concept_set_members__from_csets_and_members_to_db(): Useful abstraction for '0 members' wrok. Re-using in regular pipeline.
    - Rename: resolve_fetch_failures: resolve_fetch_failures_excess_items. This is to disambiguate from new script resolve_fetch_failures_0_members
    - Update: resolve_fetch_failures_excess_items: To more clearly state its purpose. Removed references of 0 members from that script.
    - Bugfix: Believe bug is fixed now where resolve_fetch_failures_excess_items() would always run during the refresh.
    - Rename: insert_fetch_status() -> insert_fetch_statuses() for clarity
    - Update: fetch failures GH action call: Now calls immediately upon failure as opposed to before when it was called once at the end of the fetch function.
    - Update: refresh_termhub_core_cset_derived_tables(): Now does not run if it is currently marked as active, and waits until inactive to start.

    General
    - Update: get_objs_by_composite_key() and get_obj_by_composite_key(): Now quotes column names, because some of our columns are keywords such as 'table' and 'primary_key'.

### DIFF
--- a/backend/db/resolve_fetch_failures_0_members.py
+++ b/backend/db/resolve_fetch_failures_0_members.py
@@ -2,46 +2,125 @@
 resulting in initial fetch of concept set members being 0."""
 import os
 import sys
+import time
 from argparse import ArgumentParser
+from datetime import datetime
+from typing import Dict, List
+
 
 DB_DIR = os.path.dirname(os.path.realpath(__file__))
-BACKEND_DIR = os.path.join(DB_DIR, '..')
-PROJECT_ROOT = os.path.join(BACKEND_DIR, '..')
+BACKEND_DIR = os.path.join(DB_DIR, "..")
+PROJECT_ROOT = os.path.join(BACKEND_DIR, "..")
 sys.path.insert(0, str(PROJECT_ROOT))
-from backend.db.utils import SCHEMA, fetch_status_set_success, get_db_connection, insert_from_dicts, \
-    refresh_termhub_core_cset_derived_tables, \
-    select_failed_fetches
+from enclave_wrangler.objects_api import concept_set_members__from_csets_and_members_to_db, \
+    fetch_cset_and_member_objects
+from backend.db.utils import SCHEMA, fetch_status_set_success, get_db_connection, select_failed_fetches, \
+    refresh_termhub_core_cset_derived_tables
 
-DESC = 'Resolve any failures resulting from fetching data from the Enclave\'s objects API.'
+DESC = "Resolve any failures resulting from fetching data from the Enclave's objects API."
 
-def resolve_fetch_failures_0_members(version_id: int = None, use_local_db=False):
+
+def _report_success(
+    cset_ids: List[int], fetch_audit_row_lookup: Dict[str, Dict], comment_addition: str = None, use_local_db=False
+):
+    """Report success for a list of concept set IDs.
+    todo: kind of weird to pass a lookup. Maybe do that beforehand and pass row dict"""
+    success_rows = []
+    for cset_id in cset_ids:
+        row = fetch_audit_row_lookup[str(cset_id)]
+        if comment_addition:
+            row['comment'] = row['comment'] + '; ' + comment_addition
+        success_rows.append(row)
+    fetch_status_set_success(success_rows, use_local_db)
+
+def resolve_fetch_failures_0_members(
+    version_id: int = None, use_local_db=False, polling_interval_seconds=30, schema=SCHEMA
+):
     """Resolve situations where we tried to fetch data from the Enclave, but failed due to the concept set being too new
     resulting in initial fetch of concept set members being 0.
     :param version_id: Optional concept set version ID to resolve. If not provided, will check database for flagged
-    failures."""
-    print('Resolving fetch failures: new concept set / 0 members')
-    print('id:')
-    print(version_id)
-    if version_id:
-        print(f'Fetching version ID: {version_id}')
-        with get_db_connection(local=use_local_db) as con:
-            pass
-            # refresh_termhub_core_cset_derived_tables(con, SCHEMA)
+    failures.
+    todo: Performance: Fetch only members, ideally: Even though we are fetching 'members', adding to the cset members
+     table requires cset version and container metadata, and the function that does this expects them to be formaatted
+     as objects, not as they come from our DB. We can fetch from DB and then convert to objects, but a lot of work for
+     small performance gain."""
+    print("Resolving fetch failures: non-draft, ostensibly new concept sets with >0 expressions but 0 members")
+    # Collect failures
+    failures: List[Dict] = select_failed_fetches(use_local_db)
+    failure_lookup: Dict[str, Dict] = {x['primary_key']: x for x in failures}
+    failed_cset_ids: List[int] = [version_id] if version_id else [
+        int(x['primary_key']) for x in failures if x['status_initially'] == 'fail-0-members']
+    failed_cset_ids = list(set(failed_cset_ids))  # dedupe
+    if not failed_cset_ids:
+        print("No failures to resolve.")
+        return
+
+    # Resolve
+    i = 0
+    t0 = datetime.now()
+    print(f"Fetching concept set versions and their related objects: {', '.join([str(x) for x in failed_cset_ids])}")
+    while len(failed_cset_ids) > 0 and (datetime.now() - t0).total_seconds() < 2 * 60 * 60:  # 2 hours
+        i += 1
+        print(f"- attempt {i}: fetching members for {len(failed_cset_ids)} concept set versions")
+        # Fetch data
+        csets_and_members: Dict[str, List[Dict]] = fetch_cset_and_member_objects(
+            codeset_ids=failed_cset_ids, handle_issues=False)
+        # - filter success & track results
+        csets_and_members['OMOPConceptSet'] = [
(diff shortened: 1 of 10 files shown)


backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -20,8 +20,6 @@ def resolve_fetch_failures_0_members(version_id: int = None, use_local_db=False)
     :param version_id: Optional concept set version ID to resolve. If not provided, will check database for flagged
     failures."""
     print('Resolving fetch failures: new concept set / 0 members')
-    print('id:')
-    print(version_id)
     if version_id:
         print(f'Fetching version ID: {version_id}')
         with get_db_connection(local=use_local_db) as con:
```
joeflack4, line 25: The following things were done in a commit directly made to `develop`, so I'm linking to those comments here.

## Set up of GitHub action
https://github.com/jhu-bids/TermHub/commit/29e8bcbe83d71f03e601e16a36597a4bff4a9353#r122809428

## GitHub action `version_id` param
https://github.com/jhu-bids/TermHub/commit/29e8bcbe83d71f03e601e16a36597a4bff4a9353#r122809466

## `call_github_action()` updates
https://github.com/jhu-bids/TermHub/commit/29e8bcbe83d71f03e601e16a36597a4bff4a9353#r122809553

## New: `call_github_action_with_params()`
https://github.com/jhu-bids/TermHub/commit/29e8bcbe83d71f03e601e16a36597a4bff4a9353#r122809481

backend/db/utils.py
```diff
@@ -271,9 +271,9 @@ def get_objs_by_composite_key(con, table: str, keys: List[str], objs: List[Dict]
     todo: refactor: look to insert_from_dicts() more secure way to do this by passing `params` to run_sql()`
     :return: dictionary with keys as the primary key and values as the row contents"""
     key_vals = {key: [str(obj[key]) for obj in objs] for key in keys}
+    conditions = ' AND '.join([f'"{k}" {sql_in(v, True)}' for k, v in key_vals.items()])
```
joeflack4, line 274: ### Quoting of keywords used in column names (`primary_key`, `table`, etc)
@Sigfried I made this change for `get_obj_by_composite_key()` and `get_objs_by_composite_key()` for the purposes of this PR, but this doesn't give us full coverage. I guess theoretically we should always be quoting the column names for these generalized functions.

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -2,46 +2,124 @@
...
+    if not failed_cset_ids:
+        print("No failures to resolve.")
+        return
+
+    # Resolve
+    t0 = datetime.now()
+    print(f"Fetching concept set versions and their related objects: {', '.join([str(x) for x in failed_cset_ids])}")
+    with get_db_connection(schema='', local=use_local_db) as con:
+        i = 0
+        while len(failed_cset_ids) > 0 and (datetime.now() - t0).total_seconds() < 2 * 60 * 60:  # 2 hours
```
joeflack4, line 61: ## Handling repeated fetches for duration
@Sigfried I know that you suggested using the existing functionality of `make_object_request()`. I was going to include that for some of these cases. However, I felt it a good idea to set the script up so that it can handle multiple failure cases (any that show up in the `fetch_audit` table), even though you can call it on a single cset. So because of multiple possible failures, I didn't want to use `make_object_request()` because that would stay stuck on 1 cset at a time.

frontend/src/pages/AboutPage.js
```diff
@@ -145,10 +145,10 @@ function AboutPage() {
 
       <TextH1>Database Refresh</TextH1>
       <TextBody>Will refresh the database with the latest data from the N3C Enclave.</TextBody>
-      <TextBody><b>IMPORTANT:</b> Concept set members are currently slowly updated in the API. As a tentative solution,
-        to prevent bugs, new containers and code sets will not be imported into TermHub until those members are also
-        available for fetching. This unfortunately slows down fetching of new code sets from being otherwise
-        instantaneous to hours or days.</TextBody>
+      <TextBody><b>IMPORTANT:</b> There is a delay in the Enclave where when a concept set draft is finalized, its
```
joeflack4, line 148: ## About page updates
@Sigfried (CC @stephanieshong @hlehmann17)
This pull request handles the issue where, when a new concept set version is finalized, its concept set members are not available for up to 45 minutes. I added a note about this for reference on the About page next to the refresh button.

I'd also like to eventually display additional info, such as if a refresh is active, how long it's been going for, and if any special new csets are being fetched. I created an issue for that:
- #498

backend/db/utils.py
```diff
@@ -83,6 +81,33 @@ def refresh_termhub_core_cset_derived_tables(con: Connection, schema: str):
...
+# todo: move this somewhere else, possibly load.py or db_refresh.py
+# todo: what to do if this process fails? any way to roll back? should we?
+# todo: currently has no way of passing 'local' down to db status var funcs
+def refresh_termhub_core_cset_derived_tables(con: Connection, schema: str, polling_interval_seconds: int = 30):
+    """Refresh TermHub core cset derived tables: wrapper function
+
+    Handles simultaneous requests and try/except for worker function: refresh_termhub_core_cset_derived_tables_exec()"""
+    i = 0
+    t0 = datetime.now()
+    while True:
```
joeflack4, line 93: ## `derived_tables_refresh_status` & concurrent requests
@Sigfried `refresh_termhub_core_cset_derived_tables()` now has additional logic to handle simultaneous requests to refresh the derived tables. 

We already have logic for this in the DB refresh. But for handling fetch failures of items/members, there's specific scripts; we're not calling the whole refresh. So I didn't want to re-use the existing `refresh_status` feature.

I also implemented a different strategy. The DB refresh handles it like 'a' below, whereas the derived tables function handles like 'b':
a. 1st refresh/script runs. 2nd one sees that 1st one is active, sets a variable, then exists. 1st one at the end checks to see if a new simultaneous request comes in, and calls itself to handle that.
b. 1st refresh/script runs and completes, without knowledge of any 2nd simultaneous request. 2nd one sees 1st is active, and waits until it is completed, then starts after that.

enclave_wrangler/objects_api.py
```diff
@@ -368,46 +369,49 @@ def fetch_cset_and_member_objects(
...
+            insert_fetch_statuses([{
+                'table': 'code_sets', 'primary_key': version_id, 'status_initially': 'fail-0-members',
+                'comment': f"Fetched 0 members after fetching {len(cset['expression_items'])} items."}])
+            call_github_action('resolve_fetch_failures_0_members.yml', {'version_id': version_id})
+
         cset_versions_with_concepts.append(cset)
         expression_items.extend(cset['expression_items'])
         member_items.extend(cset['member_items'])
 
-    if insert_fetch_status:
```
joeflack4, line 404: ## Failures handled immediately when happen
@Sigfried You wanted me to change things so that when we find an issue during the refresh, we immediately handle it, rather than waiting until the end of `fetch_cset_and_member_objects()` to do that.

Initially I disagreed with you, because I thought that we'd get several instances where we'd have failures due to >10k expression items, etc, and that we'd have a bunch of failures queued up waiting for us to download a dataset to fix that, and I thought it'd be better to do this all at once.

In reality, that is a very rare issue. So I changed to immediately upon fetch failure.

I suppose I could have the different failure types (excessive items, vs 0 members) handled differently in this way, but it really doesn't matter.

joeflack4:
@Sigfried I just ran this through [GitHub action](https://github.com/jhu-bids/TermHub/actions/runs/5709336967/job/15468025356) and at this point it is still running.

It looks so far to be successful. Here were the outstanding failure cases before:
before

And after running this for a few minutes, it was able to add concept set members for these 2 failure cases and remake the derived tables, and set success timestamp with updated comment:
after

There is still 1 outstanding problematic cset (`337213998`), and the script is still running. I anticipate it will run for the full pre-programmed 2 hours before it finally exits out and considers this cset to also be resolved, sheerly out of it taking so long that it simply must be OK (this is what you and I decided). Note that I brought up a discussion about this 1 strange cset on Slack. If we think that it should actually have >0 concept set members, we should probably bring it up to Amin et al. See: [slack discussion](https://cd2h.slack.com/archives/C047FEPM05U/p1690674802931519).

Issue #484, opened 2023-07-13
Unit tests: database integrity


joeflack4:
## Overview
We probably need a good set of unit tests to run on the database to check integrity. They can be ran after the refresh from datasets, refresh from objects, regularly, or ad hoc.

Will need to think through this, but basically we should run a set of tests on every table/view.

joeflack4:
@Siggie: I'm not sure how this is related to:
#396

joeflack4:
Closing #501 just to do cleanup, but it is still likely the best start if we continue this work.

PR #501, opened 2023-08-04
Test/DB Refresh: refresh & database integrity

hugwuoke:
This is what I'm currently thinking, it basically follows my understanding of the the issue #484, 

It is missing 
1. Checking in with the frontend
2. Having the ddl tables because I wasn't sure what they were
3. I was unsure if using the output from fetch_cset_and_members was the best way of comparing because from my understanding fetch_cset_and_members gets it's information from the api and not through the database as it doesn't use SQL.


test/test_database_integrity/test_database_integrity
```diff
@@ -0,0 +1,372 @@
+import os
```
joeflack4, line 1: @hugwuoke I'm going to review and merge this some time soon. If I merge it before you pick up on it again, we can continue any work needed in a new PR.

Only thing I noticed here at a glance is that `test_database_integrity` needs the file extension `.py` at the end!

joeflack4:
Closing just for cleanup but keeping issue open:
- #484

PR #502, opened 2023-08-06
DB refresh: Vocab and counts tables

joeflack4:
Updates
```
    DB refresh: Vocab and counts tables
    - Add: refresh_voc_and_counts.py
    - Add: refresh_voc_and_counts.yml

    DB refresh: general
    - Update: 0-members script: Added a todo
    - Add: DDL for concept_ancestor_plus
    - Bugfix: DDL for concept_relationship_plus: An extra comma was causing syntax error

    Enclave Wrangler
    - Update: Renamed 'FAVORITE_DATASETS' to 'DATASET_REGISTRY' and did the same for similarly named variables.
    - Update: Refactored download_datasets() to be easier to understand, less error prone, and more debuggable.t status
    - Update: Refactored get_last_update_vocab() to get_last_update_of_dataset(): This function is now general purpose.
    - Delete: get_last_update_vocab(): No longer needed. General purpose one was better fit for use in current case.
    - Update: refresh_termhub_core_cset_derived_tables_exec() -> refresh_derived_tables(): After creating CORE_CSET_DDL_MODULES and passing it in, this general purpose function now serve several different cases, as well as exhibit the exact same behavior for core derived tables. Now also knows automatically which views it needs to recreate, based on new VIEWS config constant and checkign that against the ddl_modules param.
    - Add: refresh_any_dependent_tables(): This is now called more directly rather than calling refresh_derived_tables(). Instead of passing the list of dervied tables you want refreshed, you pass the list of tables updated, and it figures out the derived tables to update in the correct order. Also added further utility functions to aid with this: get_dependent_tables_queue() and extract_keys_from_nested_dict().
    - Add: config.py: CORE_CSET_DDL_MODULES, VIEWS: Used by refresh_derived_tables(), and now these are in a convenient place with other configuration about database tables.

    Backend
    - Update: initialize(): Added missing 'relationship' table to DATASET_TABLES, utilized by load()
    - Add: to backend/db/config.py: DERIVED_TABLE_DEPENDENCIES: Maps out what tables are dependent o
n what. Can use this for a variety of different refresh situations to figure out which tables need u
pdating.
```

### DIFF
--- a/.github/workflows/refresh_voc_and_counts.yml
+++ b/.github/workflows/refresh_voc_and_counts.yml
@@ -3,7 +3,7 @@ name: Refresh vocabulary and counts tables
 on:
   schedule:
 #    - cron: '0 0 * * 0'  # weekly, sunday 12am
-#    - cron: '0 6 * * *'  # every day 6am GMT (1-2am EST/EDT)
+    - cron: '0 6 * * *'  # every day 6am GMT (1-2am EST/EDT)
   workflow_dispatch:
 jobs:
   refresh-voc-and-counts:
--- a/backend/db/config.py
+++ b/backend/db/config.py
@@ -1,5 +1,7 @@
 """Configuration for TermHub database"""
 import os
+from typing import Dict, List
+
 from dotenv import load_dotenv
 
 DB_DIR = os.path.dirname(os.path.realpath(__file__))
@@ -35,6 +37,39 @@ CONFIG_LOCAL = {
     'port': os.getenv('TERMHUB_DB_PORT_LOCAL'),
     'personal_access_token': os.getenv('GH_LIMITED_PERSONAL_ACCESS_TOKEN')
 }
+
+def invert_list_dict(d1: Dict[str, List[str]]) -> Dict[str, List[str]]:
+    """Invert a dictionary with lists as values"""
+    d2 = {}
+    for key, values in d1.items():
+        for value in values:
+            if value in d2:
+                d2[value].append(key)
+            else:
+                d2[value] = [key]
+    return d2
+
+
+def recursify_key_in_list_dict(d1: Dict[str, List[str]], key: str) -> Dict:
+    """Given a dictionary with keys and values that may or may not be keys in the dictionary, and given a specific key,
+    recursively go through and build a dictionary for that key of all of the recursive key value mappings."""
+    if key in d1:
+        inner_map = {}
+        for value in d1[key]:
+            inner_map[value] = recursify_key_in_list_dict(d1, value)
+        # return {key: inner_map}
+        return inner_map
+    else:
+        return {}
+
+
+def recursify_list_dict(d1: Dict[str, List[str]]) -> Dict[str, Dict]:
+    """Given a dictionary with keys and values that may or may not be keys in the dictionary, recursively go through and
+     build a dictionary of each key as it maps to its values."""
+    d2 = {}
+    for key in d1.keys():
+        d2[key] = recursify_key_in_list_dict(d1, key)
+    return d2
 
 def get_pg_connect_url(local=False):
     """Get URL to connect to the database server"""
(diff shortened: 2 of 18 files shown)


.github/workflows/refresh_voc_and_counts.yml
```diff
@@ -0,0 +1,44 @@
+name: Refresh vocabulary and counts tables
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'  # weekly, sunday 12am
```
joeflack4, line 5: @Sigfried I set this to run Sundays at 12am. Let me know if you can think of a better time.

backend/db/refresh_voc_and_counts.py
```diff
@@ -0,0 +1,47 @@
+"""Refresh vocabulary and counts tables
```
joeflack4, line 1: FYI: The python script that does the work.

enclave_wrangler/datasets.py
```diff
@@ -342,21 +342,27 @@ def transform(fav: dict) -> pd.DataFrame:
     return df
 
 
-def get_last_vocab_update():
-    """
-    https://unite.nih.gov/workspace/documentation/developer/api/catalog/services/CatalogService/endpoints/getTransaction
-    https://unite.nih.gov/workspace/documentation/developer/api/catalog/objects/com.palantir.foundry.catalog.api.transactions.Transaction
-    """
-    dataset_rid = FAVORITE_DATASETS['concept']['rid']
+def get_last_update_of_datasets(rid: str):
```
joeflack4, line 345: I simplified `get_last_vocab_update ()` and abstracted logic from it into new `get_last_update_of_dataset()`, which I will also be using to check when counts tables were updated, as I assume they _might_ not be updated at the same time.

joeflack4: ### Vocab and counts update time differs
```py
    x1 = get_last_update_of_dataset('concept_set_counts_clamped')  # '2023-08-03T23:46:00.951Z'
    x2 = get_last_update_of_dataset('deidentified_term_usage_by_domain_clamped')  # '2023-08-03T22:31:09.607Z'
    x3 = get_last_vocab_update()  # '2023-08-05T15:15:52.706Z
```
Q1: Do counts tables get updated at roughly the same time than vocab tables?
No. The last time the counts were updated, they were updated 2 days earlier, or rather, I would assume 5 days after the vocab tables were previously updated.
Q2: Does each count table get updated at roughly the same time?
Yes. The timestamps differ by about an hour and a half, but that is not so significant. I imagine they are updated via the same scheduled pipeline.

joeflack4: @Sigfried Also, what do you think that "clamped" means?

enclave_wrangler/config.py
```diff
@@ -145,14 +145,14 @@
         'rid': 'ri.foundry.main.dataset.f945409a-37f1-402f-a840-29b6bd675cb0',
         'column_names': ["codeset_id", "approx_distinct_person_count", "approx_total_record_count"],
         'sort_idx': ['codeset_id'],
-        'dataset_groups': ['vocab']
+        'dataset_groups': ['counts']
```
joeflack4, line 148: Because the counts tables and the vocab tables [are updated days apart](https://github.com/jhu-bids/TermHub/pull/502#discussion_r1285261401), it seems most advisable to run this [cronjob daily](https://github.com/jhu-bids/TermHub/pull/502#discussion_r1285140413), rather than weekly, and have it only do anything when it sees if counts or vocab tables are outdated.

enclave_wrangler/config.py
```diff
@@ -77,7 +77,7 @@
 # todo: All indexes should be listed here, and initialize/refresh refactored such that indexex get created by generating
 #  SQL from here, rather than reading from the DDL files.
 # Ordered because of transformation dependencies
-FAVORITE_DATASETS = OrderedDict({
```
joeflack4, line 80: @Sigfried Finally renamed `FAVORITE_DATASETS` to `DATASET_REGISTRY`. Makes more sense now I think. We now register information about a number of datasets, and rather than downloading all of our 'favorite' ones at once, we typically select them based on groupings now.

backend/db/load.py
```diff
@@ -8,22 +8,26 @@
...
 
 DB = CONFIG["db"]
 SCHEMA = CONFIG['schema']
 DATASET_TABLES = [
-    'code_sets',
+    # Vocab
     'concept',
     'concept_ancestor',
     'concept_relationship',
+    'relationship',
```
joeflack4, line 22: @Sigfried This table was missing in `initialize()`. I feel like it should be there.

.github/workflows/refresh_voc_and_counts.yml
```diff
@@ -0,0 +1,45 @@
+name: Refresh vocabulary and counts tables
```
joeflack4, line 1: @Sigfried `modified:   termhub-csets (modified content)`: Do you want me to include these updated files I downloaded in this PR?

backend/db/config.py
```diff
@@ -75,3 +75,52 @@ def get_pg_connect_url(local=False):
 if CONFIG["server"] == 'mysql':
     BRAND_NEW_DB_URL = BRAND_NEW_DB_URL + '?charset=utf8mb4'
 DB_URL = BRAND_NEW_DB_URL.replace(f'{CONFIG["port"]}', f'{CONFIG["port"]}/{CONFIG["db"]}')
+# DERIVED_TABLE_DEPENDENCIES: Shows which tables are needed to create a derived table. Generally the idea is that when
+# the dependency tables are updated, the dependent table also needs to be updated. But some tables in here have
+# dependencies but do not meet that use case. for example, 'concept_set_members_with_dups' depends on
+# 'concept_set_members', but that table is temporary.
+# todo: are the 'module names' part of the DDL file names the same as the keys in this dictionary? If so, write that
+#  down here as a comment. Could be useful for dynamically selecting the DDL files.
+DERIVED_TABLE_DEPENDENCIES = {
```
joeflack4, line 84: ### Recursively determining which derived tables need updating
@Sigfried I think this is the best way to do updating of derived tables. Basically, this is a list of tables, and the tables they depend on. So now that we have multiple kinds of refreshes, it will be useful to list out the tables that were just updated (e.g. 4 tables in the case of vocab), and then from `DERIVED_TABLE_DEPENDENCIES` we can recursively find all of the dependent derived tables, so we know exactly what to update.

This in theory is better than hard-coding the derived tables that need updating in the correct order, which is what I've done so far with the refresh.

I think this will make this PR take a little longer though.

joeflack4: @Sigfried Just wanted to ask you about `omopconceptset` and `omopconceptsetcontainer`. Are we still using these tables? How are they created? From the objects API? What are they used for as opposed to `code_sets` and `concept_set_container`?

backend/db/utils.py
```diff
@@ -35,27 +36,12 @@
 SCHEMA = CONFIG["schema"]
 
 
-def refresh_termhub_core_cset_derived_tables_exec(con: Connection, schema: str):
+def refresh_derived_tables(con: Connection, ddl_modules=CORE_CSET_DDL_MODULES, schema=SCHEMA):
```
joeflack4, line 39: ### General purpose `refresh_derived_tables()`
Previously `refresh_termhub_core_cset_derived_tables()`, this new function is general purpose and can refresh any set of derived tables, so long as you pass it `ddl_modules`.

`ddl_modules` was a static list we had to keep up-to-date in this function. Eventually I want to have it auto-computed based on the new `RECURSIVE_DEPENDENT_TABLE_MAP`. But for now I've renamed it to `CORE_CSET_DDL_MODULES` and moved it into `backend/db/config.py`.

joeflack4: ### `refresh_any_dependent_tables()`
This is a layer on top of `refresh_derived_tables()`. This 
This is now called more directly rather than calling refresh_derived_tables(). Instead of passing the list of dervied tables you want refreshed, you pass the list of tables updated, and it figures out the derived tables to update in the correct order. Also added further utility functions to aid with this: get_dependent_tables_queue() and extract_keys_from_nested_dict().

backend/db/ddl-14-concepts_with_counts_ungrouped.jinja.sql
```diff
@@ -17,5 +17,5 @@ SELECT DISTINCT
 FROM {{schema}}concept c
 LEFT JOIN {{schema}}deidentified_term_usage_by_domain_clamped tu ON c.concept_id = tu.concept_id);
 
-CREATE INDEX ccu_idx1{{optional_index_suffix}} ON {{schema}}concepts_with_counts_ungrouped(concept_id);
+CREATE INDEX ccu_idx1{{optional_index_suffix}} ON {{schema}}concepts_with_counts_ungrouped{{optional_suffix}}(concept_id);
```
joeflack4, line 20: Had to update some DDL files with `{{optional_suffix}}` to get working w/ the refreshing.

docs/backend/db/analysis.md
```diff
@@ -1,66 +1,66 @@
 # DB row counts
```
joeflack4, line 1: ## Negative deltas
@Sigfried I did not expect this. This is only for 3/4 of the vocab tables (the 4th one, `relationship`, is new). Both counts tables had positive deltas though.

Do you think we should ask Amin or someone else in-the-know why they would think that this might be the case?

Screen Shot 2023-08-08 at 11 42 31 PM

Screen Shot 2023-08-08 at 11 43 04 PM

backend/db/ddl-17-concept_ancestor_plus.jinja.sql
```diff
@@ -0,0 +1,23 @@
+-- Table: concept_ancestor_plus ------------------------------------------------------------------------------------
```
joeflack4, line 1: @Sigfried Put new `concept_ancestor_plus` in its own DDL. Added some missing things in there like `{{schema}}`, `{{optional_suffix}}`, and `{{optional_index_suffix}}`.

backend/db/ddl-16-concept_relationship_plus.jinja.sql
```diff
@@ -6,58 +6,35 @@
...
-DROP TABLE IF EXISTS {{schema}}concept_relationship_plus CASCADE;
+DROP TABLE IF EXISTS {{schema}}concept_relationship_plus{{optional_suffix}} CASCADE;
 
-CREATE TABLE IF NOT EXISTS {{schema}}concept_relationship_plus AS (
+CREATE TABLE IF NOT EXISTS {{schema}}concept_relationship_plus{{optional_suffix}} AS (
   SELECT    c1.vocabulary_id AS vocabulary_id_1
           , cr.concept_id_1
           , c1.concept_name AS concept_name_1
           , c1.concept_code
-          , c1.total_cnt AS total_cnt_1,
```
joeflack4, line 16: @Sigfried Bugfix; was an extra `,`
- Added some missing `{{optional_suffix}}` to `concept_relationship_plus` DDL

joeflack4: Bugfix: Reran and there was another extra `,`

joeflack4: Bugfix: Re-ran, encountered, and fixed:
`sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateColumn) column "total_cnt_1" specified more than once`

PR #507, opened 2023-08-11
Refresh fix(es)

joeflack4:
### Updates
Bugfix: Csets not showing up after refresh
- Update: Added a buffer period of 2 days to refreshes. This should address the problem, though we s
till have not located the root of the issue, whether it comes from TermHub or the Enclave.

Additional updates
- Update: GH action 'fetch failures - 0 members' now runs daily 7am GMT
- Update: Refresh exit message to be more clear
- Update: Increased verbosity of fetch_cset_and_member_objects()
- Bugfix: handle_response_error(): was erroring on response.content because was calling it like a function.


.github/workflows/refresh_db.yml
```diff
@@ -2,8 +2,8 @@ name: DB Refresh (Objects API)
 
 on:
   schedule:
-#    - cron: '*/20 * * * *'  # every 20 minutes
     - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
+#    - cron: '*/20 * * * *'  # every 20 minutes
```
joeflack4, line 6: ### "Fetch failures - 0 members": Now 1x/day
It is not strictly necessary that it runs 1x/day but it is a good failsafe, especially for now.

joeflack4: ### "Fetch failures - 0 members": It looked like was broken, but don't think it was
**Why I thought it was broken:**
I was surprised to see that there was 1 unresolved case in `fetch_audit`: `121711287 (Hope Termhub Test (v5))`

**Why I think it's not broken:**
I checked DB refresh nightly logs for the last 7 days (it actually was not broken; just some days had no new csets), this is the only one that did not appear.

This indicates that hope fetched this using the refresh button on dev or prod.

And I'm thinking that when she clicked that button, that was the version of the codebase where it did add it to the `fetch_audit` log, but before my "fetch failures - 0 members" was implemented. So it didn't get resolved.

---

Given this, it does not indicate that it is strictly necessary to have "fetch failures - 0 members" run nightly, but I think (and I think you agree) that we should activate it nightly as a failsafe, especially for now, so I've done that.

joeflack4, line 6: ### DB refresh cronjob _was_ working
By working, I mean that it wasn't simply exiting out with "refresh done; nothing to do" for no reason.

When we spoke earlier, I looked at several days of logs, and for those days, there actually were _no new csets created_; I confirmed.

I checked a 7 day period, and the results were: 
- stephanie's cset: should have found it, but didn't. problematic
- a 4 day period where there were *no* new csets, so it was correct when it said it had nothing to do
- before that, i checked the dates of csets created, and how many csets were gotten back from the refresh that day, and it got back the correct amount each day

joeflack4, line 6: ## ~25% of new csets not being fetched
That being said, this was a bit alarming. I'm glad we've decided on this buffer idea, as that should solve this until we hopefully locate the root issue.

When I was doing this refresh for the last 30 days, I checked to see which csets were already in our DB, and I found that only 100 of 131 of them were previously there.

**Ramifications**
1. Until the root of the issue is found, we need that time buffer on `since`.
2. Especially if this is an enclave issue, we can't reliably expect csets to be available after clicking the refresh button.

**My current plans**
I have set up a procedure that I will run, and set up reminders every 2 weeks. I'll check every 2 weeks for some time (maybe up to 6+ weeks), and make sure that 100% of the new csets are indeed being fetched.

backend/db/refresh.py
```diff
@@ -57,6 +59,7 @@ def refresh_db(
         if since and dp.parse(since) > dp.parse(last_refresh) and not force_non_contiguity:
             raise ValueError(SINCE_ERR)
         since = since if since else last_refresh
+        since = dp.parse(since) - timedelta(hours=buffer_hours)
```
joeflack4, line 62: ### 2 day buffer
I've set a buffer of 48 hours now. I agree it makes sense for the buffer to be longer for the cronjob, but with my limited time, it's easier/fine to do it this way for now. And I'll be checking every 2 weeks as well to make sure things are working as expected.

.github/workflows/refresh_db.yml
joeflack4, line 6: ### GH action > backend directly
This experience reinforces my position that we should probably change the refresh from calling the backend directly back to using the GitHub action. As it applies to this current situation:

1. I don't have (reasonable) access to logs
I want to see not just the cronjobs, but also logs from when refresh button was used, and I can't see that in GitHub now.
2. The refresh button won't have fixes/updates until next deployment
I've added the 2 day buffer, for example, but the refresh button will still be using an old version of the codebase. And we ought not rush things and implement a new deployment just for that.

backend/db/refresh.py
joeflack4, line 62: ### PSA regarding debugging and `refresh_status` = `active`
When I was troubleshooting this earlier, the refresh status was `active`. I don't know if that's because your local process was still running, or someone had clicked the button within the last few minutes, or maybe you exited out of your debugger without letting it complete. But yeah, if you exit debugging without letting it complete, it will still remain `active`. So we have to set it to `inactive` manually.

Maybe you already knew this. I can't remember now if I needed to set it manually to `inactive` earlier or if it resolved after a few minutes.

joeflack4, line 62: ### PSA about API staiblity
I did actually get a `503 Service Unavailable Error` earlier today, FYI.

I also was doing a run and it was stuck for an hour in `requests.request()`. Worked fine when I restarted the process.

PR #533, opened 2023-08-21
Bugfix: 'fetch_failures_0_members' script

joeflack4:
- Bugfix: 'fetch_failures_0_members' script was not running. The problem was that it was silent failing due to incorrect data type on version_id param; needed to be string.

### DIFF
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -403,7 +403,7 @@ def fetch_cset_and_member_objects(
             insert_fetch_statuses([{
                 'table': 'code_sets', 'primary_key': version_id, 'status_initially': 'fail-0-members',
                 'comment': f"Fetched 0 members after fetching {len(cset['expression_items'])} items."}])
-            call_github_action('resolve_fetch_failures_0_members.yml', {'version_id': version_id})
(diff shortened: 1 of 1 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -403,7 +403,7 @@ def fetch_cset_and_member_objects(
             insert_fetch_statuses([{
                 'table': 'code_sets', 'primary_key': version_id, 'status_initially': 'fail-0-members',
                 'comment': f"Fetched 0 members after fetching {len(cset['expression_items'])} items."}])
-            call_github_action('resolve_fetch_failures_0_members.yml', {'version_id': version_id})
+            call_github_action('resolve_fetch_failures_0_members.yml', {'version_id': str(version_id)})
```
joeflack4, line 406: The bugfix

joeflack4 (commented):
@Sigfried Fixed this bug. We need to re-deploy to get this fix in though.

I tested this locally, IDK how this bug slipped through. Sorry about that. We can capture this in unit testing.

We need a longer term solution to handle situations where the GitHub action fails. Should we error out? Produce / log warnings? We could kick off a GitHub action (assuming it succeeds as well) that will basically act as an email and let us know about these issues.

PR #541, opened 2023-08-30
Bugfix: Backend perpetually down after connection error

joeflack4:
- Bugfix: PendingRollbackError: When some error happened in our deployment, the DB connection was stuck in a permanent 'pending rollback' state. However, there was no way to roll back without re-deploying. The basic change needed here was that connection objects were being re-used, and were not set up in such a way that they could handle these types of issues. Changed every instance where this was happening to instead set up a new connection if no connection object was passed, rather than initializing the connection within the function definition. This was the source of the problem. The connection object being created within the definition gets created on the first call and is re-used on subsequent calls.

### DIFF
--- a/backend/db/queries.py
+++ b/backend/db/queries.py
@@ -2,11 +2,14 @@
 from functools import cache
 from typing import List, Dict
 from fastapi import Query
+from sqlalchemy import Connection
+
 from backend.db.utils import sql_query, sql_query_single_col, get_db_connection, sql_in
 
 
-def get_concepts(concept_ids: List[int], con=get_db_connection(), table:str='concepts_with_counts') -> List:
+def get_concepts(concept_ids: List[int], con: Connection = None, table:str='concepts_with_counts') -> List:
     """Get information about concept sets the user has selected"""
+    con = con if con else get_db_connection()
     q = f"""
           SELECT *
           FROM {table}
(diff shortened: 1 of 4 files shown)


backend/db/queries.py
```diff
@@ -2,11 +2,14 @@
...
 from fastapi import Query
+from sqlalchemy import Connection
+
 from backend.db.utils import sql_query, sql_query_single_col, get_db_connection, sql_in
 
 
-def get_concepts(concept_ids: List[int], con=get_db_connection(), table:str='concepts_with_counts') -> List:
+def get_concepts(concept_ids: List[int], con: Connection = None, table:str='concepts_with_counts') -> List:
     """Get information about concept sets the user has selected"""
+    con = con if con else get_db_connection()
```
joeflack4, line 12: I just made this same page in a number of functions:
Delete: `con=get_db_connection()`
Add: ` con: Connection = None` and ` con = con if con else get_db_connection()`

joeflack4:
Deployed to dev and prod. Tested and appears to be working now. Merging.

joeflack4:
The fix for this is explained in the [top comment / OP (original post)](https://github.com/jhu-bids/TermHub/pull/541#issue-1872730918).

Here's a further example illustrating the issue:

```py
from backend.db.utils import get_db_connection

con = get_db_connection()
con2 = get_db_connection()

diff = con is con2  # Falsedef conny(con=get_db_connection()):
    return con

x1 = conny()
x2 = conny()

diff2 = x1 is x2  # True
```

Because the connection crashed the first time the function was called, all subsequent calls were still using the same broken connection.

Sigfried:
Wow, I didn't know that! Thanks for fixing.

PR #555, opened 2023-09-13
Not all csets are being fetched

joeflack4:
## Updates
QC: Missing csets
Utilizing unit test test_csets_match_enclave() to create a GitHub action that will check on this daily and alert us if fails.

`48fddd85d5555c9e63462581f92c891202e3aa74`
 - Add: GitHub action

`f4cb256390a8c00654aec5d4782d7df947fffb16`
- Add: GitHub action
- Update: test_database.py: Codestyle changes, minor optimizations, broke tests into two

General
- Bugfix: Fixed a unit test import error
- Add: find_and_add_missing_csets_to_db(), add_missing_csets_to_db(), find_termhub_missing_csets(), find_bidirectional_missing_csets()
- Update: csets_and_members_enclave_to_db(): Now accepts codeset_ids as an optional parameter as opposed to 'since'. Used by newly added functions to find/add missing csets.
- Update: objects_api.py: Revived CLI, adding some of the above functionality as commands.


enclave_wrangler/objects_api.py
```diff
@@ -288,11 +289,17 @@ def all_new_objects_enclave_to_db(since: Union[datetime, str]) -> Dict[str, List
     return objects
 
 
-def csets_and_members_enclave_to_db(con: Connection, since: Union[datetime, str], schema=SCHEMA) -> bool:
+def csets_and_members_enclave_to_db(
```
joeflack4, line 292: This function now accepts either `since` or `cset_ids`

enclave_wrangler/objects_api.py
```diff
@@ -892,16 +941,20 @@ def cli():
     #     del kwargs_dict['download_favorite_objects']
     #     run(**kwargs_dict)
 
+    # New args added from 9/12/2023
+    parser.add_argument(
+        '-A', '--find-and-add-missing-csets-to-db', action='store_true',
```
joeflack4, line 946: ### CLI
Reactivated `objects_api` CLI w/ 2 commands:
- `--find-and-add-missing-csets-to-db`
- `--find-missing-csets`

I updated the DB counts, then ran `--find-and-add-missing-csets-to-db`, and then ran the counts again.

We can use this command in the future if we need to do this again.

test/test_database.py
```diff
@@ -3,48 +3,44 @@
...
 
 class TestDatabaseCurrent(unittest.TestCase):
+    """Tests for database"""
+
+    @classmethod
+    def setUpClass(cls):
+        """Fetch cset IDs to used by further tests"""
+        cls.db_codeset_ids, cls.enclave_codeset_ids = get_bidirectional_csets_sets()
+
+    def test_all_enclave_csets_in_termhub(self):
```
joeflack4, line 30: ## Unit test refactor
I did a little refactor. I split up into 2 different tests. The GitHub action will only run `test_all_enclave_csets_in_termhub()`.

When these 2 tests need to run, there's a `setUpClass()` function that will run first so that it only needs to fetch counts 1 time. This method calls `get_bidirectional_csets_sets()`, which is now one of several new functions in `objects_api.py`.

.github/workflows/refresh_db.yml
```diff
@@ -44,11 +44,21 @@ jobs:
...
+      # todo: Temporarily leaving this commented out here until we verify that root cause  of missing csets(
+      #   https://github.com/jhu-bids/TermHub/issues/573) has been resolved (likely through
+      #   https://github.com/jhu-bids/TermHub/issues/574)
+#      - name: Fetch missing csets
+#        run: make fetch-missing-csets
+
+      # Test
+      # todo: add more tests, ideally
+      - name: Test missing csets
+        run: make test-missing-csets
```
joeflack4, line 64: ## Db refresh test
Rather than making a separate GH action to run this test, I simply run this test at the end of the DB refresh.

test/test_database.py
```diff
@@ -12,12 +12,16 @@
...
 PROJECT_ROOT = TEST_DIR.parent
 sys.path.insert(0, str(PROJECT_ROOT))
-from enclave_wrangler.objects_api import find_and_add_missing_csets_to_db, get_bidirectional_csets_sets
+from enclave_wrangler.objects_api import get_bidirectional_csets_sets
 
-# TODO: temp
-# Note: Ad hoc import: On 2023/09/12, we found 93 missing csets. The below code was used to import them.
-if __name__ == '__main__':
-    find_and_add_missing_csets_to_db()
+FAIL_MSG = (
```
joeflack4, line 17: ## Test failure message
If test fails, gives some idea as to why it might have failed, what csets were missing, and how to fix the issue in the interim.

PR #576, opened 2023-09-28
 Bug: Frontend test GH action broken

joeflack4:
## Updates
Bugfix: Frontend tests GitHub action: It was using local webserver config which we are not requiring for this action at this time (satisfied with testing dev and prod).

### DIFF
--- a/.github/workflows/tests_and_qc_frontend.yml
+++ b/.github/workflows/tests_and_qc_frontend.yml
@@ -1,10 +1,8 @@
-# TODO: get to run in GH action (right now times out), then reactivate "on"
-# TODO: (a) start backend within GH action, or (b) use dev's backend URL (won't always work)
-# TODO: expose params to workflow_dispatch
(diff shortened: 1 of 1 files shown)


.github/workflows/tests_and_qc_frontend.yml
```diff
@@ -20,14 +18,14 @@ jobs:
...
-#      run: cd frontend
     - name: Install dependencies
-#      run: cd frontend && npm ci
+#      run: cd frontend && npm ci  # only playwright is necessary to install
       run: cd frontend && npm install playwright
     - name: Install Playwright browsers
       run: cd frontend && npx playwright install --with-deps
+    - name: Re-configure for non-local web server
+      run: |
+        cd frontend && sed -n '/webServer: \[/,/\]/!p' playwright.config.js > tmp && mv tmp playwright.config.js
```
joeflack4, line 28: The bugfix: I forgot I needed to remove this for the GH action. It tries to set up local sever. We eventually do want to test local inside the action, but there are other barriers to that. So for now we just want to test dev/prod, so we need to remove this local webServer config before running playwright.

PR #577, opened 2023-09-29
DB Refresh: Bug: Voc/Counts GH action - not running && test/fix indexes/pk's

joeflack4:
## Changes
`85061db3cf7740f61f0da384ce1b3c684397a14b`
Bugfix: Counts and Vocab refresh action(s)
- Bugfix: Restored chunking of parquet file downloads. Removing this ended up corrupting some parquet files.
- Bugfix: Unique index creation
- Add: load_dataset_group() abstraction, mainly to help with clean testing
- Delete: refresh_voc_and_counts.yml: now they run separately
- Update: Cronjob time for voc/counts actions: Now one runs 2 hours after the other as a safety, although they could probably run simultaneously.
- Add: tests for checking current indexes/pkeys integrity on current tables in dataset group
- Update: GitHub actions now run test at end

`77a1338070c8d38b24f86ebdb8b0dd9324e53542`
Bugfix: Counts and Vocab refresh action(s)
- Bugfix: Counts and Vocab refresh GH action: running out of memory. This was solved by removing a few unnecessary steps from our download process, where we thought we had to combine the parquet files and save them to CSV. Then, because we had a dynamic "chunking" operation which we weren't using in order to somehow save on memory, we needed to re-read the saved parquet file(s) and we were concatenating them in a single CSV. The end result was that we were using a lot of needless memory. Now, instead we are downloading the parquet files, converting to CSV on the fly for each part of the CSV, and appending to the output file.
- Update: To use premium GitHub 'large runners' to solve resources issue
- Bugfix: Now deletes old, backed up versions of remade tables at the end. Couldn't drop because some derived tables were still dependent.
- Bugfix: Vocab tables were being filtered based on if the concept or cset IDs appeared in other tables. This is no longer helpful (and is actually bad) now that we refresh via objects API.
- Bugfix: Was not reporting dataset group as having been updated at end.
- Bugfix: Wasn't doing counts at end

Update: Generalized refresh of dataset groups
- Update: refresh_voc_and_counts.py -> refresh_dataset_group_tables.py: Generalized this script so that we can have more flexibility in which groups we run.
- Update: GH action: Also updated the GitHub action so that it runs these groups sequentially. That way, at least one group may successfully complete.
- Update: makefile: With goals for each action

More related updates
- Update: Dataset group refresh stability: Now this task, similar to the normal DB refresh, will check to see if there's currently a dervied table refresh active and, if so, wait until it is done before proceeding. This will help in the incident of multiple refreshes happing simultaneously, especially now since we are splitting the vocab and counts refreshes apart.

General
- Update: makefile: Added missing phony targets, and removed some unnecessary stuff.


joeflack4:
@Sigfried Progress update

It's not disk space that is the issue. The GH action runners allow up to 25GB~ per job. We're using about 5~.

It is a memory issue. Maximum allowed is 6GB. It spikes above this. At least one time, I saw it was at 8GB (this is probably around the highest it can get). However, I'm having a bit of difficulty. I wonder if you have experience memory profiling and might have any ideas? The main problem I'm having is that I'm at a point where I think the overhead should be almost 0, but the process is still using up 4.5GB, and I don't know what's using up all that memory. I checked all the frames and variables, `locals()` and `globals()`. I only have one idea; it might be that a file buffer is not closing as I would expect, so I will at least check that. But if I can't reduce this 4.5GB, then I am loathe to try to optimize here, because that only gives me 1.5GB to work with before I run over the 6GB limit of the action. The alternative to optimization would be to pay for some premium GH action runners. I'm not sure how much that costs yet.

backend/db/utils.py
```diff
@@ -214,12 +214,6 @@ def set_search_path(dbapi_connection, connection_record):
     return engine.connect()
 
 
-def chunk_list(input_list: List, chunk_size) -> List[List]:
```
joeflack4, line 217: Your suggestion worked. Got rid of a lot of old stuff we didn't need now.

enclave_wrangler/datasets.py
```diff
@@ -37,15 +33,15 @@
 #     #'content-type': 'application/json'
 # }
 DEBUG = False
-# TODO: Once git LFS set up, dl directly to datasets folder, or put these in raw/ and move csvs_repaired to datasets/
 CSV_DOWNLOAD_DIR = os.path.join(TERMHUB_CSETS_DIR, 'datasets', 'downloads')
 CSV_TRANSFORM_DIR = os.path.join(TERMHUB_CSETS_DIR, 'datasets', 'prepped_files')
 DESC = 'Tool for working w/ the Palantir Foundry enclave API. This part is for downloading enclave datasets.'
-os.makedirs(CSV_TRANSFORM_DIR, exist_ok=True)
+for d in [CSV_DOWNLOAD_DIR, CSV_TRANSFORM_DIR]:
+    os.makedirs(d, exist_ok=True)
```
joeflack4, line 40: This was a bug I had to fix. The download dir didn't exist, so needed to be created. Kind of confusing, because I went to look back at the GH action log from when I first ran this, and it somehow ran successfully several times before.

.github/workflows/refresh_counts.yml
```diff
@@ -41,5 +41,4 @@ jobs:
           pip install -r requirements.txt
 
       - name: Refresh counts
-        run: |
-          python backend/db/refresh_dataset_group_tables.py --dataset-group counts
+        run: make refresh-counts
```
joeflack4, line 44: ## Split up tables
@Sigfried @Leskena2 I split up the "refresh vocabulary and counts tables" into 2 actions "refresh vocab" and "refresh counts" as Lisa wisely suggested.

## Still having trouble getting either to run
### Counts
Failing for a strange new reason. [Unable to read parquet](https://github.com/jhu-bids/TermHub/actions/runs/6476787183/job/17586030329).
FYI I also tried to see if I could read a parquet file as bytes directly from the `response`. It seems like pandas supports that, but I couldn't get it to work.

### Vocab
Failing for same reason. Getting some message which is not telling me why it is failing. Assume it has something to do with memory or disk space, but from what I can see so far, neither limitation is met. I also heard there may be some kind of CPU limitation, but I have no way to check what that limitation might be, assuming what I've heard is correct. I can continue troubleshooting, but may be in our best interest to pay for premium GitHub runners. I could be wrong, but preliminary estimate is $12/mon.

## Keeping `refresh_voc_and_counts.yml` for now
@Sigfried We have a lot of actions, and you wanted to cull the number of them, but I'm loathe to do that, as (1) it is also useful to have many actions, and (2) it allows us to go back and visit logs of these old actions.
I realize that '2' is not a good reason to keep long term, so I'd like to keep the combined `refresh_voc_and_counts.yml` action temporarily at least until these 2 new actions (vocab and counts) are running in stable fashion.

enclave_wrangler/datasets.py
```diff
@@ -191,62 +131,23 @@ def transform_dataset__concept_relationship(dataset_name: str) -> pd.DataFrame:
...
 def transform_dataset__concept_ancestor(dataset_name: str) -> pd.DataFrame:
     """Transformations to concept_ancestor.csv"""
     df = pd.read_csv(os.path.join(CSV_DOWNLOAD_DIR, dataset_name + '.csv'), keep_default_na=False).fillna('')
     df = transforms_common(df, dataset_name)
-
-    # JOIN
-    try:
-        csm_df = pd.read_csv(
-            os.path.join(CSV_TRANSFORM_DIR, 'concept_set_members.csv'), keep_default_na=False).fillna('')
-        ca_csm = df.merge(csm_df, left_on='ancestor_concept_id', right_on='concept_id')
```
joeflack4, line 243: @Sigfried This is where my vocab refresh was failing. There was this transform function that needed `concept_set_members.csv` as a dependency, but it was not pulled down through Git LFS, nor downloaded.

However I am pretty sure we don't want to do this operation anymore. Basically, several of these vocab tables were getting filtered based on the code set and concept IDs that appear in the `concept_set_members` and `code_sets` tables. I think we were doing this in order to decrease the size of these files. Perhaps this was important pre-Postgres, and this still made sense when we were refreshing via the datasets API. But now that we're refreshing via objects API, it does not make sense to do this filtering operation. So I have removed this filtering code from several functions.

I'm re-running the action now, and I think that should fix it.

backend/db/refresh_dataset_group_tables.py
```diff
@@ -64,14 +64,20 @@ def refresh_voc_and_counts(skip_downloads: bool = False, schema=SCHEMA):
             for statement in statements:
                 run_sql(con, statement)
             # print('Recreating derived tables')  # printed w/in refresh_derived_tables()
-            refresh_any_dependent_tables(con, config['tables'])
-
+            refresh_derived_tables(con, config['tables'], schema)
+            print('Deleting old, temporarily backed up versions of tables')
```
joeflack4, line 68: ## Bugfix: "Can't drop tables because dependencies"
Needed to move this to after remaking derived tables. No idea how this script was able to succeed before without this step being at the end.

test/test_backend/db/test_refresh_dataset_group_tables.py
```diff
@@ -0,0 +1,59 @@
...
+THIS_INPUT_DIR = THIS_DIR / 'input' / 'test_refresh_dataset_group_tables'
+VOC_INPUT_DIR = THIS_INPUT_DIR / 'test_vocab'
+PROJECT_ROOT = THIS_DIR.parent.parent.parent
+DOWNLOAD_DIR = PROJECT_ROOT / 'termhub-csets' / 'datasets' / 'prepped_files'
+sys.path.insert(0, str(PROJECT_ROOT))
+TEST_SCHEMA = 'test_n3c'
+
+
+# todo: could create some more test cases (i) too many expressions, (ii) >100k members, (iii) 0 members?
+class TestRefreshDatasetGroupTables(unittest.TestCase):
```
joeflack4, line 26: ## Automated tests
@Sigfried I created this because I have no idea why the indexes are not being created when refreshing vocab / counts tables. This is a good practice anyway, though.

backend/db/refresh_dataset_group_tables.py
```diff
@@ -0,0 +1,127 @@
...
+                        if 'does not exist' not in str(err) and alternate_dataset_dir:
+                            raise err
+
+                print('Recreating derived tables')  # derived tables
+                refresh_derived_tables(con, config['tables'], schema)
+                print('Deleting old, temporarily backed up versions of tables')
+                for table in config['tables']:
+                    run_sql(con, f'DROP TABLE IF EXISTS {schema}.{table}_old;')
+            except Exception as err:
+                print('Error occurred during table refresh. Resetting tables to pre-refresh state; restoring backups.',
```
joeflack4, line 95: ## Resetting tables when error occurs
@Sigfried I think it would be nice if I had the time to set this up so that this exception handling occurred for all types of refreshes.

I've been wanting to do this for a while. Basically it is hella annoying, when I'm troubleshooting and in development, naturally errors occur, then I have to manually go and delete all the `_new` tables, and rename all the `_old` tables back to their original states. Finally automated that here.

test/utils.py
```diff
@@ -0,0 +1,52 @@
...
+            tables_by_schema[schema] = list_tables(con, schema, filter_temp_refresh_tables=True)
+        missing_test_tables = set(tables_by_schema['n3c']) - set(tables_by_schema['test_n3c'])
+        # Populate
+        con = get_db_connection(schema='')
+        run_sql(con, f"""
+            DO $$ 
+            DECLARE
+                table_name text;
+            BEGIN
+                FOREACH table_name IN ARRAY {tuple(missing_test_tables)}
```
joeflack4, line 38: ## Populate missing tables in `test_n3c`
@Sigfried This will be helpful for several of current and upcoming tests. Sometimes we add a new table, and it is something that is needed in order for one of the DB refresh related tests to work. I've created a function that will run some SQL which will add any missing tables into `test_n3c`, and will populate with the first 50 rows from the same table in `n3c`.

There are a couple different test cases that can use this, so I've subclassed those test cases so that whenever they run, it detects automatically if anything is missing in `test_n3c` and adds.

backend/db/utils.py
```diff
@@ -688,4 +665,26 @@ def delete_codesets_from_db(codeset_ids):
     with get_db_connection() as con:
         code_sets_to_be_deleted = sql_query(
             con, f"""SELECT codeset_id, concept_set_name FROM code_sets WHERE id IN ({codeset_ids.sql_format()})"""
-        )
\ No newline at end of file
+        )
+
+
+def reset_temp_refresh_tables(schema: str = SCHEMA):
```
joeflack4, line 671: ## Resetting tables when error occurs
@Sigfried I think it would be nice if I had the time to set this up so that this exception handling occurred for all types of refreshes.

I've been wanting to do this for a while. Basically it is hella annoying, when I'm troubleshooting and in development, naturally errors occur, then I have to manually go and delete all the `_new` tables, and rename all the `_old` tables back to their original states. Finally automated that here.

backend/db/refresh_dataset_group_tables.py
joeflack4: I have a new implementation, so closing this comment.

test/test_backend/db/tst_config.py
```diff
@@ -0,0 +1,48 @@
...
+# noinspection DuplicatedCode
+THIS_DIR = Path(os.path.dirname(__file__))
+PROJECT_ROOT = THIS_DIR.parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from backend.db.config import DERIVED_TABLE_DEPENDENCY_MAP, STANDALONE_TABLES
+from backend.db.utils import show_tables
+
+
+class TestBackendDbConfig:
```
joeflack4, line 23: ## `TestBackendDbConfig`
@Sigfried I added a new, important test to make sure that our database table/view dependency configuration is up to date.

I noticed there were some new tables in the DB, but they were not in this configuration. It's important that we remember to always do the following things:
1. If there's a new derived table/view, it need to be added to `DERIVED_TABLE_DEPENDENCY_MAP`
2. If there's a new table/view that is standalone and not derived, and no table/view is derived from it, it needs to be added to `STANDALONE_TABLES`
3. If a table's DDL has changed / it has changed so that it is now dependent on different tables from before, this must be updated to reflect in `DERIVED_TABLE_DEPENDENCY_MAP`

For case (1), I ran the test and found these two new tables:
- `concept_graph_with_rel`: I don't see DDL for it or references in the codebase. @Sigfried just confirming this is an ad hoc and standalone table?
- `cset_term_usage_rec_counts`: I see this is a new table that depends on other tables. I will add update the config. Also FYI this needs to be put in its own DDL.

My test also strangely found this table, but I don't see it in the `n3c` schema! @Sigfried What's up with this one?
- `api_call_group_id_seq`

For case (2), I updated `STANDALONE_TABLES` to reflect:
- 'api_call_group_id_seq', - @Sigfried Is this correct. Standalone? No derived / anything dependent on?
- 'junk',
- 'relationship',
- 'researcher'

For case (3), this one worries me a little. There's not an easy way to make sure that `DERIVED_TABLE_DEPENDENCY_MAP` remains updated. I suppose one really complicated thing would be to write a test that compares the current DDL against the state of the DDL at the time the test last ran successfully, and do some elaborate text parsing to determine if there were dependency changes. If I can do this, then `DERIVED_TABLE_DEPENDENCY_MAP` could also theoretically be created automatically, and we could have an action to update that once in a while, perhaps bundled in with the refresh. Would be really awesome. But it sounds very strange and error-prone.
I did notice that `all_csets` dependencies are not updated. One of the new tables is a dependency. I'll update the config. But I wonder if you can tell me if anything else has changed recently off top of your head.

joeflack4: Moved this to new PR

backend/db/ddl-11-all_csets.jinja.sql
```diff
@@ -1,6 +1,7 @@
 -- Table: all_csets ----------------------------------------------------------------------------------------------------
 DROP TABLE IF EXISTS {{schema}}all_csets{{optional_suffix}} CASCADE;
 
+DROP TABLE IF EXISTS {{schema}}cset_term_usage_rec_counts;
```
joeflack4, line 4: ## Bugfix: `relation "cset_term_usage_rec_counts" already exists`
I think this happened because table was created and process exited before it could be deleted. So I added this line right before table creation as a fallback.

backend/db/ddl-6-cset_members_items.jinja.sql
```diff
@@ -41,8 +41,8 @@ JOIN {{schema}}concept c ON COALESCE(csm.concept_id, item.concept_id) = c.concep
 WHERE csm.codeset_id IS NOT NULL
    OR item.codeset_id IS NOT NULL;
 
-CREATE INDEX {{optional_index_suffix}}csmi_idx1 ON {{schema}}cset_members_items{{optional_suffix}}(codeset_id);
+CREATE INDEX csmi_idx1{{optional_index_suffix}} ON {{schema}}cset_members_items{{optional_suffix}}(codeset_id);
```
joeflack4, line 44: ## Fix: Suffix position reversed
In `cset_members_items` DDL. Fixed just for internal consistency / linguistic correctness.

backend/db/ddl-19-concept_graph.jinja.sql
```diff
@@ -1,6 +1,6 @@
-DROP TABLE IF EXISTS concept_graph CASCADE;
+DROP TABLE IF EXISTS {{schema}}concept_graph{{optional_suffix}} CASCADE;
```
joeflack4, line 1: ## Bugfix: `concept_graph` DDL
This was created in November but I'm noticing it now as I'm continuing to make sure that the counts refresh test passes. I'm surprised this didn't come up when I manually updated counts last month.

I added:
- `{{schema}}`
- `{{optional_suffix}}`
- `{{optional_index_suffix}}`

backend/db/refresh_dataset_group_tables.py
```diff
@@ -0,0 +1,132 @@
...
+
+                # Primary keys
+                # - setting all of them; quick operation, and only creates if not exist
+                statements: List[str] = get_ddl_statements(schema, ['primary_keys'], return_type='flat')
+                for statement in statements:
+                    run_sql(con, statement)
+
+                # Indexes
+                statements: List[str] = get_ddl_statements(schema, ['indexes'], return_type='flat')
+                statements = [x for x in statements if f'{schema}.{table}(' in x]  # filter for this table
```
joeflack4, line 60: ## Performance: Indexes now created as soon as table created

backend/db/utils.py
```diff
@@ -727,7 +706,8 @@ def list_tables(con: Connection = None, schema: str = None, filter_temp_refresh_
 
 
 def get_ddl_statements(
-    schema: str = SCHEMA, modules: List[str] = None, table_suffix='', index_suffix='', return_type=['flat', 'nested'][1]
+    schema: str = SCHEMA, modules: List[str] = None, table_suffix='',return_type=['flat', 'nested'][1],
+    unique_index_names=True,
```
joeflack4, line 710: ## Bugfix: Indexes not getting created
This was only happening for the vocab/counts refreshes. I forgot to pass a unique index hash, so it was detecting that an index (from the temp table that was about to get deleted) was already there, and not creating.

As part of this fix, I did a refactor on `get_ddl_statements()`, and basically made it so that when this function is called, the unique index names automatically get created.

.github/workflows/refresh_voc.yml
```diff
@@ -3,11 +3,12 @@ name: Refresh vocabulary tables
 on:
 #  schedule:
 ##    - cron: '0 0 * * 0'  # weekly, sunday 12am
-#    - cron: '0 6 * * *'  # every day 6am GMT (1-2am EST/EDT)
+#    - cron: '0 8 * * *'  # every day 6am GMT (3-4am EST/EDT)
```
joeflack4, line 6: ## Actions updates
- Counts/voc each run separately
- Voc runs 2 hours after counts, just in case

test/test_backend/db/tst_refresh_dataset_group_tables.py
```diff
@@ -0,0 +1,116 @@
...
+        table_data: Dict = self._analyze_group_tables(group_name, schema)
+        for table, data in table_data.items():
+            # - test: Expected indexes on each table
+            self.assertEqual(len(data['indexes_expected']), len(data['indexes_and_pkeys']) - len(data['pkeys']))
+            # - test: Concept table only: Has primary key(s)
+            if table == 'concept':
+                self.assertGreater(len(data['pkeys']), 0)
+
+    def test_current_vocab(self):
+        """Test current tables setup for group (indexes and, if applicable, primary keys): vocab"""
```
joeflack4, line 100: ## Test for running in action
Added tests not only to check that loading works correctly, but also to see if indexes/pkeys are properly set on the current vocab/counts tables. This runs at the end of actions.

PR #581, opened 2023-10-09
DB refresh tweaks

joeflack4:


.github/workflows/refresh_db.yml
```diff
@@ -2,8 +2,8 @@ name: DB Refresh (Objects API)
 
 on:
   schedule:
-    - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
-#    - cron: '*/20 * * * *'  # every 20 minutes
+#    - cron: '0 5 * * *'  # every day 5am GMT (12-1am EST/EDT)
+    - cron: '*/20 * * * *'  # every 20 minutes
```
joeflack4, line 6: @Sigfried @Leskena2 I decided to set the refresh rate to to once every 20 minutes (not once every 2 hours as we talked about today, nor 1x/day as it was previously), for a few reasons:
1. GitHub doesn't seem to care if we run this often.
2. It exits if no data is found, without remaking derived tables or anything.
3. We've set up a process to where if the user does a refresh while it's running, it is not a problem.
4. If all works smoothly, this is better UX, as users will less frequently need to manually refresh and wait.

backend/db/refresh.py
```diff
@@ -97,7 +97,7 @@ def cli():
         '-l', '--use-local-db', action='store_true', default=False, required=False,
         help='Use local database instead of server.')
     parser.add_argument(
-        '-b', '--buffer-hours', default=48, required=False,
+        '-b', '--buffer-hours', default=0, required=False,  # we were defaulting to 48 for a while
```
joeflack4, line 100: @Sigfried I removed the buffer period where we were fetching 48 hours past the timestamp of the last refresh, because: during the GitHub action, after it runs the refresh, it also runs the test you set up, and will throw an error if it notices anything missing.

Removing the buffer period will make sure that the action completes much more quickly if no data is there, and won't do any unnecessary writes to the database.

joeflack4:
[Action passed](https://github.com/jhu-bids/TermHub/actions/runs/6462062331/job/17543007022). Merging.

PR #584, opened 2023-10-12
Bug: Refresh: Failing on blank container ID

joeflack4:
## Updates
- Bugfix: Unhandled case where container ID could not be fetched, and container ID was blank. This was the result of TermHub just recently gaining full access to cset dra
fts. We were fetching all drafts, including ones from when sourceApplicationVersion w
as 1.0 (we're now on 2.0), which was a different data model which had no containers.
Incompatible with TermHub, so we are filtering out and not importing these csets.

## The issue
See log: https://github.com/jhu-bids/TermHub/actions/runs/6489411749/job/17623626419
```
ValueError: Enclave API returned cset version with container of id , but failed to call data for that specific container.
```

### DIFF
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -337,8 +337,13 @@ def get_bidirectional_csets_sets(con: Connection = None) -> Tuple[Set[int], Set[
     archived_container_names = [container['properties']['conceptSetName'] for container in archived_containers]
     enclave_codesets = [
         cset for cset in enclave_codesets
-        if 'conceptSetNameOMOP' in cset.keys()
-           and not cset['conceptSetNameOMOP'] in archived_container_names]
+        if 'conceptSetNameOMOP' in cset.keys()  # why is it missing sometimes? old data model?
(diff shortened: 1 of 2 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -407,6 +407,9 @@ def fetch_cset_and_member_objects(
         cset_versions: List[Dict] = [fetch_cset_version(_id, retain_properties_nesting=True) for _id in codeset_ids]
     else:
         cset_versions: List[Dict] = fetch_objects_since_datetime('OMOPConceptSet', since, verbose)
+    # - Filter any old drafts from sourceApplicationVersion 1.0. Old data model. No containers, etc. See also:
+    #  https://github.com/jhu-bids/TermHub/actions/runs/6489411749/job/17623626419
+    cset_versions = [x for x in cset_versions if x['properties']['conceptSetNameOMOP']]
```
joeflack4, line 412: The bugfix. See explanation: https://github.com/jhu-bids/TermHub/pull/584#issue-1939046960

Issue #574, opened 2023-09-24
Bug: Perpetual "refresh in progress" can occur


joeflack4:
## Overview
It can happen sometimes that the `manage` table variable `refresh_status` will be left as `active` even after a refresh has quit. The most common reason for this (the only one I've encountered) is when doing local development/debugging, as quitting the debugger will not activate the DB refresh's exception handling, which includes a step that sets the status back to `inactive`.

The effect of this is that the refresh will be broken until this issue is noticed and fixed; it just won't run. I believe that this has been the case for almost a month, since August 26th, and is why [the counts](https://github.com/jhu-bids/TermHub/blob/develop/docs/backend/db/analysis.md) haven't changed / been regularly updated since that time.

## Sub-tasks
- [x] Fix for normal refresh
- [x] Fix for any updates to derived tables

## Proposed solution
Originally, I didn't want a `refresh_status` variable. I wanted to use the variables `last_refresh_request` and a new variable `last_refresh_exited`. And based on this, we could dynamically determine if the refresh was active. If `last_refresh_exited` > `last_refresh_request`, it is inactive. Else, it is either active, or exited without reporting as such.

I propose we make this change. And then we set a threshold. If `last_refresh_request` > `last_refresh_exited`, but it has been, say 6 hours (the maximum amount of time a GH action can run), since `last_refresh_request`, then we will infer that the refresh is currently not running, and we can proceed to do the refresh.

We could also keep the `refresh_status` variable, and simply change it to `inactive` in cases of this (6 hour) threshold being exceeded.

## Related
- #573 
- #521

joeflack4:
Re-opening, as there is another edge case with the DB variable `derived_tables_refresh_status`, which also has `active`/`inactive`, so the same thing can (and just did to me) happen.

joeflack4:
This has been fixed for a few weeks!

PR #596, opened 2023-10-29
Bug: Perpetual "refresh in progress" can occur #574

joeflack4:
- Bugfix: perpetual refresh: This could happen if the refresh exited in an abnormal way mid-refresh, such as debugging and manually hitting stop. In this case, the code that would normally set refresh status to 'inactive' would not run. Getting around this problem now by using timestamps.

### DIFF
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -13,7 +13,8 @@ PROJECT_ROOT = os.path.join(BACKEND_DIR, '..')
 sys.path.insert(0, str(PROJECT_ROOT))
 from backend.db.analysis import counts_update,counts_docs
 from backend.db.config import CONFIG
-from backend.db.utils import current_datetime, get_db_connection, last_refresh_timestamp, update_db_status_var, check_db_status_var, delete_db_status_var
+from backend.db.utils import current_datetime, get_db_connection, is_refresh_active, last_refresh_timestamp, \
+    update_db_status_var, check_db_status_var, delete_db_status_var
 from enclave_wrangler.objects_api import csets_and_members_enclave_to_db
 
 DESC = 'Refresh TermHub database w/ newest updates from the Enclave using the objects API.'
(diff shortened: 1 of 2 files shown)


backend/db/refresh.py
```diff
@@ -67,14 +67,16 @@ def refresh_db(
...
         except Exception as err:
             update_db_status_var('last_refresh_result', 'error', local)
-            update_db_status_var('refresh_status', 'inactive', local)
             print(f"Database refresh incomplete; exception occurred. Tallying counts and exiting.", file=sys.stderr)
             counts_update('DB refresh error.', schema, local, filter_temp_refresh_tables=True)
             counts_docs()
             raise err
+        finally:
+            refresh_complete_dt = current_datetime()
+            update_db_status_var('last_refresh_exited', refresh_complete_dt, local)  # TODO
```
joeflack4, line 76: With the addition of this `last_refresh_exited`, `refresh_status` is now somewhat redundant. But we can keep it if it pleases us.

backend/db/refresh.py
```diff
@@ -46,7 +47,8 @@ def refresh_db(
     local = use_local_db
     print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
     t0, t0_str = datetime.now(), current_datetime()
-    if check_db_status_var('refresh_status') == 'active':
+
+    if is_refresh_active():
```
joeflack4, line 51: This now replaces the check `if check_db_status_var('refresh_status') == 'active':`

backend/db/utils.py
```diff
@@ -261,6 +261,21 @@ def is_table_up_to_date(table_name: str, skip_if_updated_within_hours: int = Non
     last_updated_key = f'last_updated_{table_name}'
     return check_if_updated(last_updated_key, skip_if_updated_within_hours)
 
+def is_refresh_active(local=False) -> bool:
```
joeflack4, line 264: New function for determining if refresh is active based on timestamps. I added to the docstring an explanation of how this works and  what the 6 hour threshold is and why it was chosen.

PR #603, opened 2023-11-02
API logger work

Sigfried:
Some refactoring here. Oh, and forgot to mention in the commit msg that schema override is still in middleware that always runs, but by itself now. It allows us to put schema=foo in the query string of any api call and expect to get the schema we want.


backend/utils.py
```diff
@@ -243,12 +175,14 @@ def set_nested_in_dict(d: Dict, key_path: List, value: Any):
 
 def return_err_with_trace(func):
     """Handle exceptions"""
+    # @joeflack4: is this helping us? it's used for four routes. it took me a while to track down
+    #   that it needed to be async and have `await func(..)`
```
Sigfried, line 179: never mind. it is useful:
image

But we need to fix the front end to take advantage of it.

PR #608, opened 2023-11-07
Populate omopVocabVersion in code_sets and derived tables #607

joeflack4:
- Add: Scripting for populating omop vocab version
- Add: Function fetch_all_csets()

Also made some minor changes / comments.

### DIFF
--- a/backend/db/ad_hoc_omop_vocab_version.py
+++ b/backend/db/ad_hoc_omop_vocab_version.py
@@ -1,0 +1,58 @@
+"""Ad hoc script: to populate omopVocabVersion in the database
+
+Inteded for single use, 2023/11/07"""
+from datetime import datetime
+from typing import Dict, List
+
+from jinja2 import Template
+
+from backend.db.utils import get_db_connection, refresh_any_dependent_tables, run_sql
+from enclave_wrangler.objects_api import fetch_all_csets
+
+
+# COALESCE was another option. What is the advantage?
+sql_jinja = """
+UPDATE n3c.code_sets
+SET omop_vocab_version = CASE
+    {% for k, v in key_vals.items() %}
+      WHEN codeset_id = {{k}} THEN '{{v}}'
+    {% endfor %}
+END
+WHERE codeset_id IN ({{id_list}});"""
+
+
+def populate_omop_vocab_version(update_code_sets=False, update_derived_tables=False):
+    """Populate omopVocabVersion in the database.
+
+    Since this is an ad hoc script meant only to be run once, defaulted params to False after successfully running.
+    :param update_code_sets (bool): Update the code_sets table.
+    :param update_derived_tables (bool): Updates derived tables."""
+    if update_code_sets:
+        print('Fetching all csets')  # 15 seconds
+        t0 = datetime.now()
+        csets: List[Dict] = fetch_all_csets()
+        key_vals = {x['codesetId']: x['omopVocabVersion'] for x in csets}
+        t1 = datetime.now()
+        print(f' - completed in {(t1 - t0).seconds} seconds')
+
+        print('Updating code_sets')  # 2 seconds
+        query: str = Template(sql_jinja).render(
+            key_vals=key_vals,
+            id_list=', '.join([str(x) for x in key_vals.keys()]))
+        with get_db_connection() as con:
+            run_sql(con, query)
+        t2 = datetime.now()
+        print(f' - completed in {(t2 - t1).seconds} seconds')
+    if update_derived_tables:
+        # Even though 'csets_to_ignore' is a depdendent derived table which will not actually include this field and
+        # thus does not need updating, I'm going to refresh all dependent tables to future-proof this in case we need to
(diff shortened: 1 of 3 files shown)


backend/db/ad_hoc_omop_vocab_version.py
```diff
@@ -0,0 +1,58 @@
+"""Ad hoc script: to populate omopVocabVersion in the database
```
joeflack4, line 1: Ad hoc script: to populate `omopVocabVersion` in the database

Since this only needs to be used one time, should we delete the script or keep it?

backend/db/utils.py
```diff
@@ -2,6 +2,8 @@
 
 todo's
   1. Some functions could be improved using `sql_query()` `params` arg to be less prone to sql injection.
+  2. Making 'Connection' optional: Can write a wrapper function and decorate all functions that need, where all it does
```
joeflack4, line 5: ### Unrelated idea 1: Making 'Connection' optional
Can write a wrapper function and decorate all such SQL based utility functions, where all it does is `con = con if con else get_db_connection()` and then pass that to inner function.

backend/db/utils.py
```diff
@@ -647,7 +650,8 @@ def get_ddl_statements(
       1. For each table: don't do anything if these tables exist & initialized
       2. Add alters to fix data types (although, should really move this stuff to dtypes settings when creating
       dataframe that loads data into db.
-      3. I think it's inserting a second ; at the end of the last statement of a given module"""
+      3. I think it's inserting a second ; at the end of the last statement of a given module
+      4. consider throwing an error if no statements found, either here, or where func is called"""
```
joeflack4, line 654: ## Unrelated idea 2: Throw error when no DDL statements found
Siggie encountered this edge case

backend/db/ad_hoc_omop_vocab_version.py
```diff
@@ -0,0 +1,58 @@
...
+            id_list=', '.join([str(x) for x in key_vals.keys()]))
+        with get_db_connection() as con:
+            run_sql(con, query)
+        t2 = datetime.now()
+        print(f' - completed in {(t2 - t1).seconds} seconds')
+    if update_derived_tables:
+        # Even though 'csets_to_ignore' is a depdendent derived table which will not actually include this field and
+        # thus does not need updating, I'm going to refresh all dependent tables to future-proof this in case we need to
+        # repurpose / reuse.
+        print('Updating derived tables')  # 1 seconds
```
joeflack4, line 50: ## Updated derived tables
@Sigfried The only dependent tables on `code_sets` were `all_csets` and `csets_to_ignore`. I see that you already added the vocab version field to both the DDL and the table for `all_csets`. Now I've updated it so that it has the versions for each cset.

`csets_to_ignore` did not need this field

backend/db/utils.py
```diff
@@ -119,6 +121,7 @@ def refresh_any_dependent_tables(con: Connection, independent_tables: List[str]
     derived_tables: List[str] = get_dependent_tables_queue(independent_tables)
     if not derived_tables:
         print(f'No derived tables found for: {", ".join(independent_tables)}')
+        return
```
joeflack4, line 124: ## Unrelated update: return early when no derived tables found
The function `refresh_any_dependent_tables()` checks a given table (e.g. `code_sets`) and finds any tables dependent on it (e.g. `all_csets`), then proceeds to call `refresh_derived_tables()`. If it didn't find any dependent table, it would print a message like "no tables now", but it would continue onto the `refresh_derived_tables()` function. `refresh_derived_tables()` doesn't do anything if it is not passed any tables, so this was harmless, but I decided it's much better to do a return and not call it.

backend/db/ad_hoc_omop_vocab_version.py
```diff
@@ -0,0 +1,58 @@
...
+    :param update_derived_tables (bool): Updates derived tables."""
+    if update_code_sets:
+        print('Fetching all csets')  # 15 seconds
+        t0 = datetime.now()
+        csets: List[Dict] = fetch_all_csets()
+        key_vals = {x['codesetId']: x['omopVocabVersion'] for x in csets}
+        t1 = datetime.now()
+        print(f' - completed in {(t1 - t0).seconds} seconds')
+
+        print('Updating code_sets')  # 2 seconds
```
joeflack4, line 38: ## Updated `code_sets`
`code_sets` has 7,392 entries. I got 7,186 entries got back from enclave after filtering out archived cset containers and source app 1.0 csets.

So after doing this update, there are 205 csets there where `omop_vocab_version = NULL`: 
[remaining null omop_vocab_version after updating all fetchable csets.csv.zip](https://github.com/jhu-bids/TermHub/files/13291701/remaining.null.omop_vocab_version.after.updating.all.fetchable.csets.csv.zip)

I'm surprised this number is not 206 (7392 - 7186). But I imagine there's a good reason; perhaps we had already set the version for one of these csets and it has since been archived.

PR #611, opened 2023-11-13
Fixing concept hierarchy

Sigfried:
Not quite done yet. Collapse/expand is broken.   Fixed now.


backend/db/ddl-19-apirun_groups.jinja.sql
```diff
@@ -0,0 +1,50 @@
+-- View: apiruns_grouped ------------------------------------------------------------------------------------
+DROP TABLE IF EXISTS public.api_runs CASCADE;
+
+CREATE TABLE IF NOT EXISTS public.api_runs (
```
joeflack4, line 4: New table. I see!

backend/db/initialize.py
```diff
@@ -92,6 +78,7 @@ def create_database(con: Connection, schema: str):
         run_sql(con2, DDL_COUNTS)
         run_sql(con2, DDL_COUNTS_RUNS)
         run_sql(con2, DDL_FETCH_AUDIT)
+        run_sql(con2, DDL_CSET_COMPARE)
```
joeflack4, line 81: New table. I see!

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
 
 
-@router.get("/subgraph")
-async def subgraph(request: Request, id: List[int] = Query(...)):   # id is a list of concept ids
-    return await subgraph_post(request=request, id=id)
-
-
-@router.get("/hierarchy")
-def hierarchy(id: List[int] = Query(...)):   # id is a list of concept ids
+def all_paths(g: networkx.DiGraph, nodes: List[int]) -> List[List[int]]:
```
joeflack4, line 59: `all_paths()`: I get what this function does and I heard your explanation earlier. It's still not 100% in my mind how it is all glued together and what the end result looks like but I'm sure that will become clear to me in time. When this is done, I'll want to see the end result.

- [ ] @joeflack4: Gain full understanding of this and then mark resolved.

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
+    for root in roots:
+        for leaf in leaves:
+            _paths = list(nx.all_simple_paths(g, root, leaf))
+            for path in _paths:
+                if len(path) > 1:
+                    paths.append(path)
+    return paths
+
+
+def fill_in_gaps(G, nodes):
```
joeflack4, line 79: `fill_in_gaps()`: I see how this works! I know that the full set of nodes is needed but I'm not sure how the frontend uses that full set, but that's OK for now.

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
-        successors = sort_by_most_descendants(successors)
-    else:
-        return {successor: create_hierarchy(G, successor, sorted) for successor in successors}
-
-def for_testing():
-    G = nx.DiGraph()
-    G.add_edges_from([(1, 2), (1, 3), (2, 4), (2, 5), (2,3), (2, 6), (3, 6), (9,2)])
-    return G
 
+class TreeNode:
```
joeflack4, line 116: Doing this in OO way, interesting. I see how this works. Interested to see it in action and see how this ends up getting serialized to JSON.

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
+def condense_super_nodes(sg, threshhold=10):
+    super_nodes = [node for node, degree in sg.out_degree() if degree > threshhold]
+    # for node in super_nodes:
+    # sg.remove_node(node) -- n
+
+def expand_super_node(G, subgraph_nodes, super_node):
+    sg = G.subgraph(subgraph_nodes)
+
+
+def generate_csv_lines(node, level=-1):
```
joeflack4, line 148: CSV?

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
+
+    def insert(self, path):
+        if path:
+            head, *tail = path
+            if head not in self.children:
+                self.children[head] = TreeNode(head)
+            self.children[head].insert(tail)
+
+
+def condense_super_nodes(sg, threshhold=10):
```
joeflack4, line 139: I don't see where `condense_super_nodes()` and `expand_super_node()` are used yet but I guess that will be soon!

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
+#         (10, 11), (11, 12), (12, 6)
+#     ])
+#     components = G.subgraph([2,3,4,5,6,7])
+#     return G, components
+
+
+
+def generate_graph_edges():
+    with get_db_connection() as con:
+        # moving the sql to ddl-20-concept_graph.jinja.sql
```
joeflack4, line 280: ## `concept_graph` table
I don't see `ddl-20-concept_graph.jinja.sql` yet, I'm guessing that's still pending?

**Questions**
1. I'm curious to know what this table looks like and how it will be used?
2. Is it not redundant with `networkx`? Does it serve a different role?
3. I'm not seeing where this table gets populated. Has that code been written yet? Is this going to be populated as we were talking about, using the `concept_releationship` `is_a` edges?

backend/routes/graph.py
```diff
@@ -29,166 +35,308 @@ def subgraph():
...
+#         (1, 2), (2, 3), (3, 4),
+#         (1, 5), (5, 6), (6, 7),
+#         (10, 11), (11, 12), (12, 6)
+#     ])
+#     components = G.subgraph([2,3,4,5,6,7])
+#     return G, components
+
+
+
+def generate_graph_edges():
```
joeflack4, line 278: This doesn't seem to generate them; it seems to fetch that which was pre-generated. Should this be renamed to `get_graph_edges()`?

backend/db/ddl-19-apirun_groups.jinja.sql
Sigfried: Yeah...added that table and also a view based on it called `apiruns_grouped`

backend/routes/graph.py
Sigfried: There's inline documentation (generated by chatgpt, but pretty good). But to give a clearer idea of the mechanics, or at least the data shapes, I created [a new documentation file](https://github.com/jhu-bids/TermHub/blob/fixing-concept-hierarchy/docs/graph.md)

backend/routes/graph.py
Sigfried: This is not actually being used yet. It's basically the same thing as `all_paths()`, but returns a subgraph instead of a path list. When we get to visualizing these hierarchies, we won't need to repeat concepts for every parent, so we'll use this.

backend/routes/graph.py
Sigfried: i just renamed it

backend/routes/graph.py
Sigfried: Forgot to add it earlier.  It gets [populated by the ddl](https://github.com/jhu-bids/TermHub/blob/fixing-concept-hierarchy/backend/db/ddl-20-concept_graph.jinja.sql) when [this code](https://github.com/jhu-bids/TermHub/blob/5aced3941d1dc4a39e0824568938d7ed4312aaa1/backend/routes/graph.py#L278C1-L318C20) runs, which happens when you start up the backend and the networkx file is missing.

Yes, it is redundant, but runs faster this way. And, btw, starting the backend is actually faster now.

backend/routes/graph.py
Sigfried: Yes, you're write. I moved the code to the ddl

PR #634, opened 2023-11-28
Big refactor of graph stuff

Sigfried:
- It should be more correct now
  - I think it's slower
  - The first two tests succeed now, but:
  - With lots of concepts (e.g., 20,000), it takes forever and the whole backend won't respond while it's processing
    - The problem isn't in connect_nodes anymore, it's in all_paths when it's looking for simple paths to root. Those paths can get long and there can be a lot of them. Maybe a strategy of looking for predecessors would work?
  - It needs a lot more inline documentation
- I may have fixed the duplicate test thing by putting the loop through tests inside the loop through environments rather than other way around.


Sigfried:
I'm going to deploy this to dev

backend/db/ddl-20-concept_graph.jinja.sql
```diff
@@ -1,14 +1,25 @@
 DROP TABLE IF EXISTS concept_graph CASCADE;
 
 CREATE TABLE IF NOT EXISTS concept_graph AS (
-     SELECT ancestor_concept_id AS source_id, descendant_concept_id AS target_id
+     SELECT ancestor_concept_id AS source_id,
```
joeflack4, line 4: Just confirming I've looked at `concept_graph` table for the first time.

backend/utils.py
```diff
@@ -16,6 +17,12 @@
 from backend.config import CONFIG
 
 
+def powerset(iterable):
```
joeflack4, line 20: `powerset`: I don't see where this is used at the moment; fill me in if you want.
I like `chain`; don't use it that often. Haven't used `combinations` before.

frontend/tests/README.md
```diff
@@ -11,13 +11,20 @@ what optimization is being performed and tested. Can take multiple values delimi
 |---------------------|------------------------------------------------------|:------------:|:--------:|
 | `as-is` or no value | Use code without attempting any special optimization |   yes  | yes |
 | `no-cache`          | Disables caching -- in DataCache.cachePut            |   yes  | yes |
-| `no-rxext-from-api` | RxNorm Extension codes left out of backend API results. See [below](#No-RxNorm-Extension-codes)|   no   | no  |
+| `no-rxext-from-api` | RxNorm Extension codes left out of backend API results. See [below](#no-rxnorm-extension-codes)|   no   | no  |
 
 
 ## Notes
 
 ### No RxNorm Extension codes
```
joeflack4, line 19: Thanks for documenting.

PR #650, opened 2023-12-13
Timeouts

joeflack4:


backend/app.py
```diff
@@ -96,6 +96,23 @@ def read_root():
...
+async def simulate_long_running_task():
+    """Simulate"""
+    await asyncio.sleep(10)
+    return "Computation complete"
+
+@APP.get("/test-timeout")
+async def test_timeout(timeout=3):
+    """Test timeout"""
+    try:
+        result = await asyncio.wait_for(simulate_long_running_task(), timeout=timeout)
```
joeflack4, line 111: ## Working method: Using Python > `gunicorn`/`uvicorn` options
@Sigfried I tried `gunicorn` as well, but those options didn't work either.

I went back to ChatGPT and tried asking the question generally. It came up with this, and it worked the first time!

Let's just go with our middleware option.

I have to complete some of my tickets first, so if you want me to do this, it'll have to be some time in January. But I would be happy if you tackled this too.

joeflack4: ## How to get this to work for all routes (middleware)
Here's what ChatGPT said: https://chat.openai.com/share/5d1061b3-b114-42fd-a3eb-0b67adad7631

Sigfried:
@joeflack4, I tried this locally:

    REQUEST_TIMEOUT_ERROR = 10
    # from https://github.com/tiangolo/fastapi/issues/1752
    # Adding a middleware returning a 504 error if the request processing time is above a certain threshold
    @APP.middleware("http")
    async def timeout_middleware(request: Request, call_next):
        try:
            start_time = time.time()
            return await asyncio.wait_for(call_next(request), timeout=REQUEST_TIMEOUT_ERROR)

        except asyncio.TimeoutError:
            process_time = time.time() - start_time
            return JSONResponse({'detail': 'Request processing time excedeed limit',
                                 'processing_time': process_time},
                                status_code=HTTP_504_GATEWAY_TIMEOUT)

And have a job that's been going on for minutes with no timeout. Did you have some better idea about how to implement it as middleware?

joeflack4:
@Sigfried I don't, off the top of my head. This is new to me. Well, only this:
- [ ] Did you make sure that the route you were testing was `async`?

PR #654, opened 2023-12-15
Bugfix: DB refresh: no members returned

joeflack4:


enclave_wrangler/utils.py
```diff
@@ -384,7 +384,7 @@ def make_objects_request(
             raise err
         else:
             print(err)
-            return
+            return []
```
joeflack4, line 387: ## Bugfix
@Sigfried Was getting a bug ([log](https://github.com/jhu-bids/TermHub/actions/runs/7215449436/job/19659712956)):

```
    member_items.extend(cset['member_items'])
TypeError: 'NoneType' object is not iterable
```

I changed `make_objects_request()` to return `[]` instead of `None` in these cases:
- When no data
- When error and not `fail_on_error`

I'm not sure if this is the best way to resolve this. I don't think either of us do. The best way would again be a good test suite that could detect if this change causes any problems.

For now, this fixes my problem.

## Problem revealed by bugfix - Draft w/ no members
Never mind. discussed with Siggie. It was caused by change in the code yesterday.
<p>

I'll bring this up to Maya on Slack. I was getting error on cset version `325128828`. I searched the enclave and nothing showed up. I checked, and of the 20 csets I got back on my fetch, this was the only one that was a draft.

I checked further and saw that it had 4 expressions but no members. I checked and 3/4 had `includeDescendants = True` and had `4/4 isExcluded = False`, so I would expect members.

I think something changed and now we are no longer getting members backed from drafts.

</p>
</details>

PR #656, opened 2023-12-15
Bugfix: DB Refresh: no members returned (follow-ups)

joeflack4:
## Updates
- Bugfix: Yesterday when I merged, I forgot to commit this last bit of code that finishes off the fix.


backend/db/refresh.py
```diff
@@ -72,6 +72,7 @@ def refresh_db(
         except Exception as err:
             print(f"Database refresh incomplete; exception occurred. Tallying counts and exiting.", file=sys.stderr)
             update_db_status_var('last_refresh_result', 'error', local)
+            update_db_status_var('last_refresh_error_message', str(err), local)
```
joeflack4, line 75: Now logs error messages in DB. I shoulda done this before.

enclave_wrangler/objects_api.py
```diff
@@ -365,12 +365,17 @@ def get_bidirectional_csets_sets(con: Connection = None) -> Tuple[Set[int], Set[
 
 def add_missing_csets_to_db(cset_ids: List[int], con: Connection = None, schema=SCHEMA):
     """Add missing concept sets to the DB"""
-    con = con if con else get_db_connection(schema=schema)
-    csets_and_members_enclave_to_db(con, cset_ids=cset_ids, schema=schema)
+    try:
+        con = con if con else get_db_connection(schema=schema)
+        csets_and_members_enclave_to_db(con, cset_ids=cset_ids, schema=schema)
+    except Exception:
+        reset_temp_refresh_tables(schema)
```
joeflack4, line 372: `add_missing_csets_to_db`: Now does `reset_temp_refresh_tables` if something bad happens

enclave_wrangler/objects_api.py
```diff
@@ -365,12 +365,17 @@ def get_bidirectional_csets_sets(con: Connection = None) -> Tuple[Set[int], Set[
...
 
 
 def find_and_add_missing_csets_to_db(con: Connection = None, schema=SCHEMA):
     """Find and add missing concept sets to the DB"""
+    if is_refresh_active():
```
joeflack4, line 377: `find_and_add_missing_csets_to_db`: Now doesn't run if refresh is active

joeflack4:
Also checked if "fetch failures 0 members" was working correctly.

Double checked if simultaneous refresh could still occur; I thought something looked strange in the action logs.

backend/db/analysis.py
```diff
@@ -167,7 +167,7 @@ def counts_over_time(
         schema, local, from_cache=True)
 
     # Pivot
-    values = 'count' if method == 'counts_table' else 'delta'
+    values = 'counts' if method == 'counts_table' else 'delta'
```
joeflack4, line 170: ## Bugfix: Counts tables totals
There was a typo here. This is a general function but it has some split logic on counts vs deltas. However, in one place it was declaring the string `count` and later was checking for the string `counts`. The spellings were not equal, so that logic was never running.

Basically, this ended up manifesting a problem in which the tables for both counts and deltas were being treated the same, such that for both tables, the aggregated sum for that given day was what was being displayed. This is correct for deltas. However, for counts, we don't want to sum up all of the counts for all runs in a given day, we just want to check the counts for the latest run.

This bug caused no problem for deltas.

joeflack4:
Db refresh was broken today. I meant to merge this on Saturday, I think I forgot to.

PR #668, opened 2024-01-09
Graph test follow-ups

joeflack4:
## Tasks
- [x] Make JSON non-indented for bigger outputs
- [x] Save outputs of any unique commits to subfolders


test/test_backend/routes/tst_graph.py
```diff
@@ -45,57 +45,59 @@
...
+        normalized_hidden = {k: sorted(list(v)) for k, v in hidden.items()}
+        results = {
+            'sg.edges': sorted(list(sg.edges)),
+            'nodes_in_graph': sorted(list(nodes_in_graph)),
+            'preferred_concept_ids': sorted(list(preferred_concept_ids)),
+            'orphans_not_in_graph': sorted(list(orphans_not_in_graph)),
+            'hidden': normalized_hidden,
+        }
+        n_records = sum([len(v) for k, v in results.items()])
+        indent = 0 if n_records > minify_threshold else 0
```
joeflack4, line 82: ## Minification
Now does minified JSON if number of records is over 1,000.

test/test_backend/routes/tst_graph.py
```diff
@@ -45,57 +45,59 @@
...
+        with open(input_path, 'w') as f:
```
joeflack4, line 83: ## Tests failing @Sigfried 
@Sigfried These tests are now failing, since the outputs of 1get_connected_subgraph()` have changed since I initially created. Leaving this task up to you now:

- [ ] Do one of the following
  - a. If this is indicative of something being wrong with your recent changes (e.g. when I made the outputs, `get_connected_subgraph()` was correct, but it's not now) --> Fix things and rerun tests until they pass
  - b. if there is no bug in `get_connected_subgraph()` and it is currently outputting what is expected, you need to save new outputs. To do so, run `TstGraph. _create_inputs__get_connected_subgraph()`.

test/test_backend/routes/tst_graph.py
```diff
@@ -45,57 +45,59 @@
...
         for case in cases:
             for k, v in case.items():
                 if k == 'codeset_ids':
                     case[k] = v.split(',')
-        cls.test_cases = cases
+        return cases
+
+    # TODO: type hints
+    @staticmethod
+    def _save_output(
```
joeflack4, line 68: ## Saves output
Now by default saves outputs when test is run to a subfolder which is named for the given git commit hash.

> [!Warning]
> Make sure all changes are committed before running and saving outputs. Otherwise the hash will be incorrect.

PR #671, opened 2024-01-10
Bug: DB connection idling; soon no available connections

joeflack4:
## Sub-tasks
- [x] Close a lot of leaky code causing idle connections
- [x] Make test for acceptable num idle w/in week

## Changes
Bugfix: Idle connections
- Bugfix: Closed much leaky code causing idle connections to appear.
- Add: Test for checking against acceptable threshold of idle connections.

General
- Add: Some docstrings
- Update: Some typing

Other
- Bugfix: get_cset_members_items(): Was re-doing the same query, undoing special conditional operations.
- Bugfix: get_concept_ids_by_codeset_id(): Was returning wrong data type when returning empty value.
- Delete: Removed duplicate route/function omop_id_from_concept_name()

### DIFF
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -3,7 +3,7 @@
 todo's
   1. Some functions could be improved using `sql_query()` `params` arg to be less prone to sql injection.
   2. Making 'Connection' optional: Can write a wrapper function and decorate all functions that need, where all it does
-  is `con = con if con else get_db_connection()` and then pass that to inner function.
+  is `conn = con if con else get_db_connection()`, run the inner function, and then close conn if not con.
 """
 import json
 import os
@@ -514,7 +514,7 @@ def run_sql(con: Connection, command: str, params: Dict = {}) -> Any:
 
 def show_tables(con: Connection = None, print_dump=True):
     """Show tables"""
-    con = con if con else get_db_connection()
+    conn = con if con else get_db_connection()
     query = """
         SELECT n.nspname as "Schema", c.relname as "Name",
               CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 't' THEN 'TOAST table' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'partitioned table' WHEN 'I' THEN 'partitioned index' END as "Type",
@@ -529,11 +529,13 @@ def show_tables(con: Connection = None, print_dump=True):
           AND pg_catalog.pg_table_is_visible(c.oid)
         ORDER BY 1,2;
     """
-    res = sql_query(con, query, return_with_keys=False)
+    res = sql_query(conn, query, return_with_keys=False)
     if print_dump:
         print(pd.DataFrame(res))
         # print('\n'.join([', '.join(r) for r in res])) ugly
         # print(pdump(res)) doesn't work
+    if not con:
+        conn.close()
     return res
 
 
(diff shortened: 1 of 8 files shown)


backend/db/utils.py
```diff
@@ -714,3 +714,19 @@ def reset_temp_refresh_tables(schema: str = SCHEMA):
                 run_sql(con, f'DROP TABLE {schema}.{table}_new;')
             else:
                 run_sql(con, f'ALTER TABLE {schema}.{table}_new RENAME TO {table};')
+
+
+# TODO
+def get_idle_connections(last_num_days: int = 0):
```
joeflack4, line 720: ## Func for getting idle connections
WIP

test/test_backend/db/test_utils.py
```diff
@@ -59,6 +60,17 @@ def test_select_failed_fetches(self):
         self.assertEqual(len(results), self.n1 + len(self.mock_data))
 
 
+class TestIdleConnections(unittest.TestCase):
+
+    def test_idle_connections(self, threshold=10):
```
joeflack4, line 65: ## Test
For checking acceptable number of idle connections within given time frame.

WIP

backend/routes/db.py
```diff
@@ -314,49 +325,54 @@ async def get_concept_ids_by_codeset_id_post(request: Request, codeset_ids: Unio
...
     await rpt.start_rpt(request, params={'codeset_ids': codeset_ids})
 
-    with get_db_connection() as con:
-        try:
-            q = f"""
-                  SELECT csids.codeset_id, COALESCE(cibc.concept_ids, ARRAY[]::integer[]) AS concept_ids
-                  FROM (VALUES{",".join([f"({csid})" for csid in codeset_ids])}) AS csids(codeset_id)
-                  LEFT JOIN concept_ids_by_codeset_id cibc ON csids.codeset_id = cibc.codeset_id"""
+    try:
+        with get_db_connection() as con:
```
joeflack4, line 345: ## Type 1/3: `with` block
There were many instances of offending `with` blocks, where they would either (a) return within the block, or there was (b) a try/except within the block. I saw both such examples in the current log of idle connections in the database.

There were quite a lot of instances of 'a'.

backend/routes/db.py
```diff
@@ -402,7 +418,8 @@ def get_researchers(id: List[str] = Query(...), fields: Union[List[str], None] =
         FROM researcher
         WHERE "multipassId" = ANY(:id)
     """
-    res: List[RowMapping] = sql_query(get_db_connection(), query, {'id': list(id)}, return_with_keys=True)
```
joeflack4, line 405: ## Type 2/3: function call inside function call
I don't know if I saw offending cases of this in the log, but I believe these remain open as well.

backend/db/utils.py
```diff
@@ -3,7 +3,7 @@
 todo's
   1. Some functions could be improved using `sql_query()` `params` arg to be less prone to sql injection.
   2. Making 'Connection' optional: Can write a wrapper function and decorate all functions that need, where all it does
-  is `con = con if con else get_db_connection()` and then pass that to inner function.
+  is `conn = con if con else get_db_connection()`, run the inner function, and then close conn if not con.
```
joeflack4, line 6: ## Type 3/3: `conn = con if con else get_db_connection()`
I implemented this pattern basically for convenience to allow us to call SQL functions without having to pass a connection. However, I was not closing these connections.

There's a `todo` here to remind us that it'd be good to have a wrapper for such cases. There's a number of instances of this pattern now.

backend/routes/db.py
```diff
@@ -158,19 +164,18 @@ def get_cset_members_items(
...
+        FROM cset_members_items
+        WHERE codeset_id {sql_in(codeset_ids)}
+    """
     with get_db_connection() as con:
-        query = f"""
-            SELECT DISTINCT {', '.join(columns)}
-            FROM cset_members_items
-            WHERE codeset_id {sql_in(codeset_ids)}
-        """
+        # TODO: these 3 following lines don't do anything; get overridden by the next query. which query is correct?
```
joeflack4, line 173: ## Bugfix todo: `get_cset_members_items()`
@Sigfried these 3 following lines don't do anything; get overridden by the next query. which query is correct?

backend/routes/db.py
```diff
@@ -497,18 +516,6 @@ def _atlas_json_from_defs(defStr: List[Dict]) -> Dict:
     return atlas_json_from_defs(defs)
 
 
-@router.get('/omop-id-from-concept-name/{name}')
-def omop_id_from_concept_name(name):
```
joeflack4, line 501: ## Bugfix: Duplicate `omop_id_from_concept_name()` (add caching?)
@Sigfried There were 2 functions for `omop_id_from_concept_name()`, and they were exactly the same, with the exception that one of them had `@cache` on it. This uncached one was the 2nd of the two, so I'm guessing the prior one was disregarded. This means that previously this was not caching, but now it will. I think this is intended though.

joeflack4 (commented):
I looked through every instance of `get_db_connection()` in the codebase (n=37). I identified 3 different ways that we open up database connections, and issues with each of them that happen in some or all instances of those ways, and fixed them.

I also looked at the database's log of currently idle connections and corrected all those cases as well; but I think this was a redundant measure with the above.

PR #679, opened 2024-01-21
Bugfix: `reset_temp_refresh_tables()`

joeflack4:
## Overview
Fixes bug where temp table could not be dropped because of dependent views.

## Updates
    Bugfix: reset_temp_refresh_tables()
    - Bugfix: reset_temp_refresh_tables(): Would error out because could not drop table with dependent views without CASCADE.
    - Add: filter_temp_refresh_tables()
    - Add: list_views()

### DIFF
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -513,9 +513,9 @@ def run_sql(con: Connection, command: str, params: Dict = {}) -> Any:
     return q
 
 
-def show_tables(con: Connection = None, print_dump=True):
+def show_tables(con: Connection = None, print_dump=True) -> List[Row]:
     """Show tables"""
-    conn = con if con else get_db_connection()
+    conn = con if con else get_db_connection(schema=SCHEMA)
     query = """
         SELECT n.nspname as "Schema", c.relname as "Name",
               CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 't' THEN 'TOAST table' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'partitioned table' WHEN 'I' THEN 'partitioned index' END as "Type",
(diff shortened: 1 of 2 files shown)


backend/db/utils.py
```diff
@@ -703,19 +703,23 @@ def reset_temp_refresh_tables(schema: str = SCHEMA):
           file=sys.stderr)
     with get_db_connection(schema=schema) as con:
         tables: List[str] = list_tables(con, schema)
```
joeflack4, line 705: ## `list_tables()` vs `show_tables()`
2 funcs that do roughly same thing. My bad, I did `list_tables()` this 6 months after Siggie did `show_tables()`. Some differences:

`list_tables()`
- Has `filter_temp_refresh_tables` (default: `False`) param to filter tables ending w/ `_old` or `_new`
- By default, returns simple list of table names
- No option to show views
- Can select `schema`

`show_tables()`
- Shows more info, i.e. schema, table name, whether table or view, creator (e.g. 'n3c', 'all_csets', 'table', 'thadmin').
- Prints info by default
- Only `n3c` schema

I guess I should refactor to use `show_tables()` instead, and integrate some of my functionality into it.

backend/db/utils.py
```diff
@@ -540,6 +540,18 @@ def show_tables(con: Connection = None, print_dump=True):
     return res
 
 
+def list_views(con: Connection = None, filter_temp_refresh_views=False) -> List[str]:
```
joeflack4, line 543: ## `list_views()`
This was useful/necessary for me so I added it.

backend/db/utils.py
```diff
@@ -626,18 +638,23 @@ def load_csv(
         update_db_status_var(f'last_updated_{table}', str(current_datetime()), local)
 
 
-def list_tables(con: Connection, schema: str = SCHEMA, filter_temp_refresh_tables=False) -> List[str]:
+def filter_temp_refresh_tables(tables: List[str]) -> List[str]:
```
joeflack4, line 641: ## `filter_temp_refresh_tables()`
I could have just had the same code in 2 places to deal w/ this but I decided to DRY it.

backend/db/utils.py
```diff
@@ -702,21 +719,23 @@ def reset_temp_refresh_tables(schema: str = SCHEMA):
...
-            run_sql(con, f'ALTER TABLE {schema}.{table}_old RENAME TO {table};')
-
-        dangling_new_tables = [t for t in tables if t.endswith('_new')]
-        for table in dangling_new_tables:
-            table = table.replace('_new', '')
-            if table in tables:
-                run_sql(con, f'DROP TABLE {schema}.{table}_new;')
-            else:
-                run_sql(con, f'ALTER TABLE {schema}.{table}_new RENAME TO {table};')
+        for item_type, func in (('VIEW', list_views), ('TABLE', list_tables)):
```
joeflack4, line 722: ## The bugfix
I was originally going to solve this with `CASCADE`, which would've probably been fine. However, I needed to fetch view names in order to rename them `e.g. remove `_old`, and since I had the list of views, I felt it'd be safer to just remove these views first before tables, which obsoleted the need for cascade.

This ended up in a nice refactor as well.

PR #680, opened 2024-01-21
Bugfix: Derived table refresh status active indefinitely

joeflack4:
## Updates
- Bugfix: Derived table refresh status active indefinitely
- Update: Codestyle in db/utils.py

### DIFF
--- a/backend/db/config.py
+++ b/backend/db/config.py
@@ -76,6 +76,7 @@ BRAND_NEW_DB_URL = \
 if CONFIG["server"] == 'mysql':
     BRAND_NEW_DB_URL = BRAND_NEW_DB_URL + '?charset=utf8mb4'
 DB_URL = BRAND_NEW_DB_URL.replace(f'{CONFIG["port"]}', f'{CONFIG["port"]}/{CONFIG["db"]}')
+REFRESH_JOB_MAX_HRS = 6  # see is_refresh_active() for documentation on this var
 CORE_CSET_TABLES = ['code_sets', 'concept_set_container', 'concept_set_version_item', 'concept_set_members']
 CORE_CSET_DEPENDENT_TABLES = [
     'cset_members_items',
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -31,7 +31,7 @@ from sqlalchemy.sql.elements import TextClause
 from typing import Any, Dict, Tuple, Union, List
 
 from backend.db.config import CORE_CSET_DEPENDENT_TABLES, CORE_CSET_TABLES, RECURSIVE_DEPENDENT_TABLE_MAP, \
-    get_pg_connect_url, VIEWS_TO_REFRESH
+    REFRESH_JOB_MAX_HRS, get_pg_connect_url, VIEWS_TO_REFRESH
 from backend.config import CONFIG, DATASETS_PATH, OBJECTS_PATH
 from backend.utils import commify
 from enclave_wrangler.models import pkey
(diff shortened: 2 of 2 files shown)


backend/db/utils.py
```diff
@@ -177,21 +179,23 @@ def refresh_termhub_core_cset_derived_tables(con: Connection, schema=SCHEMA, pol
...
             try:
-                update_db_status_var('derived_tables_refresh_status', 'active')
+                update_db_status_var('last_derived_refresh_request', current_datetime(), local)
                 # The following two calls yield equivalent results as of 2023/08/08. I've commented out
                 #  refresh_derived_tables() in case anything goes wrong with refresh_any_dependent_tables(), since that
                 #  is based on a heuristic currently, and if anything goes wrong, we may want to switch back. -joeflack4
                 # refresh_derived_tables(con, CORE_CSET_DEPENDENT_TABLES, schema)
                 refresh_any_dependent_tables(con, CORE_CSET_TABLES, schema)
             finally:
-                update_db_status_var('derived_tables_refresh_status', 'inactive')
```
joeflack4, line 194: ## Replaced `inactive`/`active` with start/stop time
Now the derived table refresh activity monitoring is done the same way as the full database refresh. 

This fixes the problem where the refresh exits unexpectedly and the except/finally blocks do not execute, e.g. when cancelling a run in development environment, or when GH action times out.

backend/db/utils.py
```diff
@@ -282,11 +287,27 @@ def is_refresh_active(local=False) -> bool:
...
     the refresh takes."""
-    last_start = dp.parse(check_db_status_var('last_refresh_request', local))
-    last_end = dp.parse(check_db_status_var('last_refresh_exited', local))
-    considered_active = last_start > last_end
-    hours_since_last_refresh: float = (dp.parse(current_datetime()) - last_end).total_seconds() / 60 / 60
-    return considered_active and hours_since_last_refresh < 6
+    # Always check if derived tables refresh active for any case. Else check more for standard refresh status
+    key_pairs = [('last_derived_refresh_request', 'last_derived_refresh_exited')]
+    if refresh_type == 'standard':
+        key_pairs = [('last_refresh_request', 'last_refresh_exited')] + key_pairs
```
joeflack4, line 293: ## Normal refresh now won't start if derived tables are active
I changed this behavior mainly handle any unanticipated edge cases that might occur. Basically, the way it was before was that if the refresh was currently active, but the derived tables refresh were active for some reason (e.g. vocab/counts refresh or some other operation in local development), the refresh would start, and it would get to the point where it start to do the derived table refresh. Then, if it detected that the derived tables were currently being updated, it would wait up to two hours. Perhaps it was fine to keep it this way, but our refresh runs every 20 minutes, so I think it's fine to simply not start the refresh at all if the derived tables are running.

backend/db/utils.py
joeflack4, line 194: ## Delete: `derived_tables_refresh_status`
I deleted the `manage` table db variable `derived_tables_refresh_status` but left the normal `db_ refresh_status` variable. Why? We don't actually need either of these. However for the DB refresh, in order to determine manually if it is active, it's easier to check the `manage` table and see that it is marked as `active`, rather than finding the 2 timestamp variables and do a calculation.
The same could be said for `derived_tables_refresh_status`, but the fact is that there isn't nearly as often a need that I need to check this status, so it just seemed better to get rid of it.

backend/db/config.py
```diff
@@ -76,6 +76,7 @@ def get_pg_connect_url(local=False):
 if CONFIG["server"] == 'mysql':
     BRAND_NEW_DB_URL = BRAND_NEW_DB_URL + '?charset=utf8mb4'
 DB_URL = BRAND_NEW_DB_URL.replace(f'{CONFIG["port"]}', f'{CONFIG["port"]}/{CONFIG["db"]}')
+REFRESH_JOB_MAX_HRS = 6  # see is_refresh_active() for documentation on this var
```
joeflack4, line 79: ## Strange: Multiple long running refreshes
multiple long running refresh

This is related but this PR does not address this issue. I just want to raise the fact that I'm not sure why this is happening. There should only ever be 2 refresh actions going simultaneously, and the newer one should only be running for a short time. It should only get to the point at the beginning of the refresh where it sees another one is active and should quit out immediately. However in this screenshot, 2 of the refreshes were going for like an hour. I have now idea how this could happen, even after checking the logs. The logs were roughly identical, and both were doing the up-to-2 hour wait for the derived tables refresh to complete.

PR #681, opened 2024-01-22
Bugfix: Follow-up fix for derived views

joeflack4:
## Overview
Siggie (and I) tried to do a fix involving refreshing of views. However Siggie felt we either didn't need view or did not need to refresh it, and so tried to deal with that. I cleaned up some of that code, but at the time neither of us realized that this would be insufficient.

I've since fixed the error caused involving views and resetting of table status when refresh fails. I can't remember if Siggie actually wants to get rid of the view(s) or not, so I'm keeping them. I'm just getting rid of the bug that errors out during the reset.

## Changes
    Bugfix: Error on views during DB refresh table reset
    - Bugfix: A different PR recently fixed the original bug, but kept the views. But a new bug was introduced, causing the view to sometimes be recognized and sometimes not. This fixes that.

    General
    - Add: reset_refresh_state(): Resets both temporary tables and status variables
    - Add: reset_refresh_status(): Reset DB refresh active status variables to reflect non-active status

    - Delete: Obsolete view related config code

### DIFF
--- a/backend/db/config.py
+++ b/backend/db/config.py
@@ -90,19 +90,6 @@ CORE_CSET_DEPENDENT_TABLES = [
     # 'csets_to_ignore',
     'cset_members_items_plus',
 ]
-
-# todo: remove obsolete view config code. can probably remove these
-# todo: obsoleted due to list_views()
-# VIEWS = [
-#     'cset_members_items_plus',
-#     # 'csets_to_ignore',
-# ]
-# todo: temporarily deactivating VIEWS_TO_REFRESH because it causes bugs since this view also needs to be removed
-#  elsewhere in the codebase. If we choose to keep the view, these 2 lines should be deleted. Otherwise if we remove
-#  the view (from both the DB and all references in codebase), there will proably be no need for these two lines either
-# VIEWS_REFRESH_IGNORE_LIST = ['cset_members_items_plus']
-# VIEWS_TO_REFRESH = list(set(list_views()) - set(VIEWS_REFRESH_IGNORE_LIST))
-
 # DERIVED_TABLE_DEPENDENCY_MAP: Shows which tables are needed to create a derived table. Generally the idea is that when
 #  the dependency tables are updated, the dependent table also needs to be updated. But some tables in here have
 #  dependencies but do not meet that use case. for example, 'concept_set_members_with_dups' depends on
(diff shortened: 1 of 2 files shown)


backend/db/config.py
```diff
@@ -79,22 +79,17 @@ def get_pg_connect_url(local=False):
...
     'concept_ids_by_codeset_id',
     'members_items_summary',
     'codeset_counts',
     'all_csets',
+    # views
     # 'csets_to_ignore',
     'cset_members_items_plus',
 ]
-# todo: try to auto detect what is a view
-VIEWS = [
```
joeflack4, line 92: ## Delete: Views related config
This bit is either now obsolete or replaced by `list_views()`.

backend/db/utils.py
```diff
@@ -137,7 +137,7 @@ def refresh_derived_tables(
      entry will appear further down in the list."""
     temp_table_suffix = '_new'
     ddl_modules_queue = derived_tables_queue
-    views = [x for x in VIEWS_TO_REFRESH if x in ddl_modules_queue]
+    views = [x for x in list_views() if x in ddl_modules_queue]
```
joeflack4, line 140: ## The bugfix
`VIEWS_TO_REFRESH` was an empty list because we didn't care to refresh. But this created a problem where, because the way things are coded, this variable was not being read at the time of creating derived tables/views, but only when deleting them. So this resulted in the new view being created and the old one backed up, but during cleanup where old tables/views were removed, this variable was empty, so the view was not removed, and then the table it depends on could not be dropped because of lack of cascade.

backend/db/utils.py
```diff
@@ -276,6 +276,31 @@ def is_table_up_to_date(table_name: str, skip_if_updated_within_hours: int = Non
...
+        last_start: datetime = dp.parse(check_db_status_var(start_time_key, local))
+        last_end: datetime = dp.parse(check_db_status_var(end_time_key, local))
+        if last_end < last_start:
+            # arbitrarily add 1 sec so considered exited after start
+            last_end: str = (last_start + timedelta(seconds=1)).astimezone(pytz.timezone('America/New_York')).isoformat()
+            update_db_status_var(end_time_key, last_end, local)
+    update_db_status_var('refresh_status', 'inactive', local)
+
+
+def reset_refresh_state(local=False):
```
joeflack4, line 298: ## Add: `reset_refresh_state()`
Meant to only be run manually during development.

Calls: 
- `reset_temp_refresh_tables()`: Function I created last month to clean up the state of temporary tables in case of a failed refresh.
- `reset_refresh_status(local)`: New func I just added in this PR

backend/db/utils.py
```diff
@@ -276,6 +276,31 @@ def is_table_up_to_date(table_name: str, skip_if_updated_within_hours: int = Non
     return check_if_updated(last_updated_key, skip_if_updated_within_hours)
 
 
+def reset_refresh_status(local=False):
```
joeflack4, line 279: ## Add: `reset_refresh_status()`
Meant to only be run manually during development.

This resets the state of all variables in `manage` related to whether or not the DB is currently active. It was starting to become a pain to have to do this manually because I just started to introduce more such variables for the derived tables.

PR #685, opened 2024-01-23
Table dependency mapping updates

joeflack4:
## Overview
This is some stuff that I recently added to #577, but I moved here instead and did further updates. In spirit of how I've been doing things in Mondo, I've tried to make PRs more atomic. And it may be important to get some of these changes merged quickly in case #577 takes some more time.

## Changes
Update DB config & add test
- Add: TestBackendDbConfig: Checks that DB config is in sync w/ DB
- Update: backend/db/config.py: Updated to keep current w/ state of DB, and added some functionality to
work with TestBackendDbConfig.

### DIFF
--- a/backend/db/config.py
+++ b/backend/db/config.py
@@ -77,6 +77,7 @@ if CONFIG["server"] == 'mysql':
     BRAND_NEW_DB_URL = BRAND_NEW_DB_URL + '?charset=utf8mb4'
 DB_URL = BRAND_NEW_DB_URL.replace(f'{CONFIG["port"]}', f'{CONFIG["port"]}/{CONFIG["db"]}')
 REFRESH_JOB_MAX_HRS = 6  # see is_refresh_active() for documentation on this var
+# Table/View configuration: for n3c schema
 CORE_CSET_TABLES = ['code_sets', 'concept_set_container', 'concept_set_version_item', 'concept_set_members']
 CORE_CSET_DEPENDENT_TABLES = [
     # tables
@@ -90,6 +91,11 @@ CORE_CSET_DEPENDENT_TABLES = [
     # 'csets_to_ignore',
     'cset_members_items_plus',
 ]
+# STANDALONE_TABLES: Not derived from any other table, nor used to derive any other table/view. Used for QC testing.
+STANDALONE_TABLES = [
+    'junk',
+    'relationship',
+]
 # DERIVED_TABLE_DEPENDENCY_MAP: Shows which tables are needed to create a derived table. Generally the idea is that when
 #  the dependency tables are updated, the dependent table also needs to be updated. But some tables in here have
 #  dependencies but do not meet that use case. for example, 'concept_set_members_with_dups' depends on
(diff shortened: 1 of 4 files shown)


backend/db/config.py
```diff
@@ -99,17 +105,21 @@ def get_pg_connect_url(local=False):
 DERIVED_TABLE_DEPENDENCY_MAP = {
     # Dependencies figured out
     # - tables
-    'all_csets': ['code_sets', 'omopconceptset', 'concept_set_container', 'omopconceptsetcontainer', 'concept_set_counts_clamped', 'codeset_counts'],  # omopconceptset is all lowercase in DB but referred to as OMOPConceptSet in DDL and enclave object API
+    # all_csets: omopconceptset is all lowercase in DB but referred to as OMOPConceptSet in DDL and enclave object API
+    'all_csets': ['code_sets', 'omopconceptset', 'concept_set_container', 'omopconceptsetcontainer',
```
joeflack4, line 109: ## Update `all_csets` in `DERIVED_TABLE_DEPENDENCY_MAP`

backend/db/config.py
```diff
@@ -99,17 +105,21 @@ def get_pg_connect_url(local=False):
...
+                  'concept_set_counts_clamped', 'cset_term_usage_rec_counts',  'codeset_counts', 'researcher'],
     'codeset_counts': ['members_items_summary'],
     'codeset_ids_by_concept_id': ['cset_members_items'],
     'concept_ancestor_plus': ['concept_ancestor', 'concepts_with_counts'],
+    'concept_graph': ['concept_ancestor'],
```
joeflack4, line 114: ## Add `concept_graph` to `DERIVED_TABLE_DEPENDENCY_MAP`

backend/db/config.py
```diff
@@ -99,17 +105,21 @@ def get_pg_connect_url(local=False):
...
     'concept_ids_by_codeset_id': ['cset_members_items'],
     'concept_relationship_plus': ['concept_relationship', 'concepts_with_counts'],
     'concepts_with_counts': ['concepts_with_counts_ungrouped'],
     'concepts_with_counts_ungrouped': ['concept', 'deidentified_term_usage_by_domain_clamped'],
     'cset_members_items': ['concept_set_members', 'concept_set_version_item'],
+    'cset_term_usage_rec_counts': ['concept_set_members', 'concepts_with_counts'],
```
joeflack4, line 120: ## Add `cset_term_usage_rec_counts ` to `DERIVED_TABLE_DEPENDENCY_MAP`
@Sigfried Actually, I see that in the DDL this is supposed to be dropped after creating `all_csets`. Is it a mistake that this is here or are you using it? If so, are you just using it temporarily?
If using it, I need to edit the DDL for `all_csets` to remove the drop, and make a new DDL for this table.
If not using, we should delete the table and remove from this config.

backend/db/config.py
```diff
@@ -99,17 +105,21 @@ def get_pg_connect_url(local=False):
...
     'members_items_summary': ['cset_members_items'],
-    'concept_graph': ['concept_ancestor'],
```
joeflack4, line 112: ## Delete  `concept_graph` from `DERIVED_TABLE_DEPENDENCY_MAP`
Table removed.

backend/db/utils.py
```diff
@@ -565,9 +565,20 @@ def run_sql(con: Connection, command: str, params: Dict = {}) -> Any:
     return q
 
 
-def show_tables(con: Connection = None, print_dump=True) -> List[Row]:
-    """Show tables"""
-    conn = con if con else get_db_connection(schema=SCHEMA)
+# todo: should add a 'include_views' param because it is ambiguous as to whether or not this returns views. it does.
+def list_schema_objects(
```
joeflack4, line 569: ## `show_tables` Updates
- Renamed to `show_schema_objects`
- Added `schema` param
- Merged `list_tables` unique functionality into here, as below
- `list_tables` now just calls this and passes fixed params
- `list_views` now just calls this and passes fixed params
- Added filters for: views, sequences, tables, temp refresh tables/views (i.e. `*_old`, `*_new`)
- Added `names_only` param to return simple list of strings

test/test_backend/db/tst_config.py
```diff
@@ -0,0 +1,47 @@
...
+# noinspection DuplicatedCode
+THIS_DIR = Path(os.path.dirname(__file__))
+PROJECT_ROOT = THIS_DIR.parent.parent.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+
+from backend.db.config import DERIVED_TABLE_DEPENDENCY_MAP, STANDALONE_TABLES
+from backend.db.utils import list_schema_objects
+
+
+class TestBackendDbConfig(unittest.TestCase):
```
joeflack4, line 24: ## Add `TestBackendDbConfig`
Added this test and now it passes after doing these other updates. Further discussion in https://github.com/jhu-bids/TermHub/pull/577#discussion_r1461297206, although most of those concerns have now been resolved. Any remaining concerns have had GH issue created.

backend/db/config.py
```diff
@@ -90,6 +91,11 @@ def get_pg_connect_url(local=False):
     # 'csets_to_ignore',
     'cset_members_items_plus',
 ]
+# STANDALONE_TABLES: Not derived from any other table, nor used to derive any other table/view. Used for QC testing.
+STANDALONE_TABLES = [
```
joeflack4, line 95: ## Add `STANDALONE_TABLES`
This helps us keep track of what should not be in `DERIVED_TABLE_DEPENDENCY_MAP` and is used by `TestBackendDbConfig`

backend/db/config.py
joeflack4: It's important to know the answer to this question, as I've kept the table but since had to remove from the config because this was breaking the counts/vocab refresh (possibly also the main refresh but probably not).

joeflack4: We went with (a), dropping the table. It just needed to be dropped from the DB 1x. In the future, the DDL automatically drops it upon creation.

PR #689, opened 2024-01-29
Bugfix: DB refresh `all_csets` err

joeflack4:
Bugfix: DB refresh erroring out on creation of all csets
This is happening as a result of us removing temp table cset_term_usage_rec_counts.
- Update: all csets DDL: removed optional suffxi from cset_term_usage_rec_counts, both for internal consistency, and also because it is not needed. The target table of the DDL is all_csets, and that should be the only place where it is needed.
- Bugfix:


joeflack4:
This wasn't an issue actually. It only failed because I dropped the table while the DDL for `all_csets` was actually running! The very next reset was able to pick up just fine.

PR #700, opened 2024-02-11
`last_refresh_success` fix / redundancy & refresh action tweak

joeflack4:
## Changes
    - Update: last_refresh_success time now set to 1 microsecond after last_refresh_request, as redundancy in order to ensure that no new data that was added in this timeframe is missed.
    - Update: DB refresh action: Deactivated fetch-missing-csets, as various tweaks, including 2 hour re-refresh buffer, and also test-missing-csets, have made this redundant and it just slows down the refresh.

### DIFF
--- a/.github/workflows/refresh_db.yml
+++ b/.github/workflows/refresh_db.yml
@@ -52,11 +52,11 @@ jobs:
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
           commit_message: Updating documentation in docs/backend/db/analysis.md
-      # todo: This should not be necessary, but some csets still not being fetched. Some thoughts in related issues:
-      #   https://github.com/jhu-bids/TermHub/issues/573
-      #   https://github.com/jhu-bids/TermHub/issues/574
-      - name: Fetch missing csets
-        run: make fetch-missing-csets
+      # fetch-missing-csets: Deactivated because various tweaks and default re-refresh buffer solve this problem, and
+      # also test-missing-csets is there as a redundancy. Original context:
+      # https://github.com/jhu-bids/TermHub/issues/573 https://github.com/jhu-bids/TermHub/issues/574
+#      - name: Fetch missing csets
+#        run: make fetch-missing-csets
       # Test
       # todo: add more tests, ideally
       - name: Test missing csets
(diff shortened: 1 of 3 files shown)


.github/workflows/refresh_db.yml
```diff
@@ -52,11 +52,11 @@ jobs:
...
-      # todo: This should not be necessary, but some csets still not being fetched. Some thoughts in related issues:
-      #   https://github.com/jhu-bids/TermHub/issues/573
-      #   https://github.com/jhu-bids/TermHub/issues/574
-      - name: Fetch missing csets
-        run: make fetch-missing-csets
+      # fetch-missing-csets: Deactivated because various tweaks and default re-refresh buffer solve this problem, and
+      # also test-missing-csets is there as a redundancy. Original context:
+      # https://github.com/jhu-bids/TermHub/issues/573 https://github.com/jhu-bids/TermHub/issues/574
+#      - name: Fetch missing csets
+#        run: make fetch-missing-csets
```
joeflack4, line 59: ## Refresh action tweak
fetch-missing-csets: Deactivated because various tweaks and default re-refresh buffer solve this problem, and also test-missing-csets is there as a redundancy.

backend/db/refresh.py
```diff
@@ -46,14 +46,18 @@ def refresh_db(
...
 
     if is_refresh_active():
         print('INFO: Refresh already in progress. When that process completes, it will restart again. Exiting.')
-        update_db_status_var('new_request_while_refreshing', t0_str, local)
+        update_db_status_var('new_request_while_refreshing', start_time, local)
         return
+    # end_time: Even though in reality the refresh will not end 1 microsecond after the start time, we're setting it
+    # this way because this is the easiest way to make sure that future refreshes will not miss any new data that was
+    # added between when the refresh started and ended.
+    end_time: str = tz_datetime_str(dp.parse(start_time) + timedelta(microseconds=1))
```
joeflack4, line 58: ## `last_refresh_success` fix / redundancy
end_time: Even though in reality the refresh will not end 1 microsecond after the start time, we're setting it this way because this is the easiest way to make sure that future refreshes will not miss any new data that was added between when the refresh started and ended.

backend/db/utils.py
```diff
@@ -236,17 +236,22 @@ def chunk_list(input_list: List, chunk_size) -> List[List]:
         yield input_list[i:i + chunk_size]
 
 
-def current_datetime(time_zone=['UTC/GMT', 'EST/EDT'][1]) -> str:
-    """Get current datetime in ISO format as a string."""
+def tz_datetime_str(dt: datetime, time_zone=['UTC/GMT', 'EST/EDT'][1]) -> str:
```
joeflack4, line 239: ## `tz_datetime_str`
Creates timezone-fromatted datetime string. Added to DRY things up a bit.

joeflack4:
As an alternative we could perhaps not have a success endtime but have could have a boolean for whether the last refresh was a success or not.

PR #718, opened 2024-03-07
Bugfix: refresh: KeyError: 'archived'

joeflack4:
## Changes
    Bugfix: DB refresh: KeyError: 'archived'
    The cause of this was that apparently when you create concept sets in a certain way (I'm guessing via the API), some fields that are normally populated (normally required?) do not get populated, such as 'alias' and 'archived'.
    - Bugfix: Now sets 'archived' to False if not detected
    - Delete: Some comments

### DIFF
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -671,10 +671,10 @@ def concept_set_members__cset_rows_to_db(con: Connection, cset: Dict, members: L
         'codeset_id': cset['codesetId'],
         'concept_id': member['conceptId'],
         'concept_set_name': cset['conceptSetNameOMOP'],
-        'is_most_recent_version': cset['isMostRecentVersion'], # not sure: is this correct?
-        'version': cset['version'], # not sure: is this correct?
+        'is_most_recent_version': cset['isMostRecentVersion'],
+        'version': cset['version'],
         'concept_name': member['conceptName'],
-        'archived': container['archived'], # not sure: is this correct?
+        'archived': container.get('archived', False),
     } for member in members]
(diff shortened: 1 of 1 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -671,10 +671,10 @@ def concept_set_members__cset_rows_to_db(con: Connection, cset: Dict, members: L
         'codeset_id': cset['codesetId'],
         'concept_id': member['conceptId'],
         'concept_set_name': cset['conceptSetNameOMOP'],
-        'is_most_recent_version': cset['isMostRecentVersion'], # not sure: is this correct?
```
joeflack4, line 674: I deleted 3 comments that said `# not sure: is this correct?`, as I think I was doing this correctly. It's been almost a year since I've done this and we've had no problems yet, so.

enclave_wrangler/objects_api.py
```diff
@@ -671,10 +671,10 @@ def concept_set_members__cset_rows_to_db(con: Connection, cset: Dict, members: L
...
-        'version': cset['version'], # not sure: is this correct?
+        'is_most_recent_version': cset['isMostRecentVersion'],
+        'version': cset['version'],
         'concept_name': member['conceptName'],
-        'archived': container['archived'], # not sure: is this correct?
+        'archived': container.get('archived', False),
```
joeflack4, line 677: ## The bugfix
The cause of this was that apparently when you create concept sets in a certain way (I'm guessing via the API), some fields that are normally populated (normally required?) do not get populated, such as 'alias' and 'archived'.

PR #721, opened 2024-03-08
Makefile help

joeflack4:
Makefile help
- Update: Running 'make help' will now show documentation for all commands.
- Update: Updated PHONY goals to be correct
- Delete: Removed some unused commands

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,8 +1,9 @@
 SRC=backend/
 
-.PHONY: lint tags ltags test all lintall codestyle docstyle lintsrc linttest doctest doc docs code linters_all codesrc \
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -128,3 +89,57 @@ serve-frontend:
 
 serve-backend:
 	uvicorn backend.app:APP --reload
+
+# Help
+help:
```
joeflack4, line 94: ## `make help`
Running this will now show show each command and what it does.

I wanted to do this in a programmatic way; to just list out all the commands. But I couldn't find a way that would work.

PR #724, opened 2024-03-09
Resolve fetch failures: clean exception handling

joeflack4:
## Changes
    - Update: resolve_fetch_failures_0_members: Now does reset_temp_refresh_tables() if encounters exception.
    - Update: resolve_fetch_failures_excess_items:
      - Now does reset_temp_refresh_tables() if encounters exception.
      - Add: Missing param: schema

### DIFF
--- a/backend/db/resolve_fetch_failures_0_members.py
+++ b/backend/db/resolve_fetch_failures_0_members.py
@@ -16,7 +16,8 @@ PROJECT_ROOT = os.path.join(BACKEND_DIR, "..")
 sys.path.insert(0, str(PROJECT_ROOT))
 from enclave_wrangler.objects_api import concept_set_members__from_csets_and_members_to_db, \
     fetch_cset_and_member_objects
-from backend.db.utils import SCHEMA, fetch_status_set_success, get_db_connection, select_failed_fetches, \
+from backend.db.utils import SCHEMA, fetch_status_set_success, get_db_connection, reset_temp_refresh_tables, \
+    select_failed_fetches, \
     refresh_derived_tables
 
 DESC = "Resolve any failures resulting from fetching data from the Enclave's objects API."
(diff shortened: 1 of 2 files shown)


backend/db/resolve_fetch_failures_excess_items.py
```diff
@@ -12,13 +12,13 @@
...
     refresh_derived_tables, \
-    select_failed_fetches
+    reset_temp_refresh_tables, select_failed_fetches
 from enclave_wrangler.datasets import CSV_TRANSFORM_DIR, download_datasets
 from enclave_wrangler.utils import was_file_modified_within_threshold
 
 DESC = "Resolve failures due to too many expression items or members when fetching data from the Enclave's objects API."
 
-def resolve_fetch_failures_excess_items(use_local_db=False, cached_dataset_threshold_hours=0):
+def resolve_fetch_failures_excess_items(schema=SCHEMA, use_local_db=False, cached_dataset_threshold_hours=0):
```
joeflack4, line 21: ## `resolve_fetch_failures_excess_items` `schema`
Added missing param. This action is still broken though for a reason I don't remember, and apparently it doesn't need to run anyway. I don't know if we've ever had more than 1 such failure case, and it was on a TermHub test set I think.

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -76,9 +77,13 @@ def resolve_fetch_failures_0_members(
...
         if success_cases:
-            with get_db_connection(schema=schema, local=use_local_db) as con:
-                concept_set_members__from_csets_and_members_to_db(con, csets_and_members)
-                refresh_derived_tables(con)
+            try:
+                with get_db_connection(schema=schema, local=use_local_db) as con:
+                    concept_set_members__from_csets_and_members_to_db(con, csets_and_members)
+                    refresh_derived_tables(con)
+            except Exception as err:
+                reset_temp_refresh_tables(schema)
```
joeflack4, line 85: ## Discussion
This is the update.

This is prompted by the fact that I saw this error message pop up in this [fetch failures log](https://github.com/jhu-bids/TermHub/actions/runs/8212911832/job/22463708956):
`sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateTable) relation "all_csets_old" already exists`

This is unexpected; I haven't seen this error pop up in `resolve_fetch_failures` before. This update doesn't fix the problem. It only ensures a clean exit if the problem does occur. So the problem can occur again. But since it happens rarely, and since the fetch failures script will run again in a day anyway, I guess it's not a huge problem.

The primary cause that I can think of for this happening would be if this script and the normal refresh were running simultaneously. However, the way I've set up `refresh_derived_tables()`, only one of these things should run at a time. So maybe that logic is broken. Other than that, I have no idea why this would occur.

Issue #520, opened 2023-08-16
New Feature: Drafts


joeflack4:
## Overview
We would like to be able to see and manipulate concept set drafts.

## Sub-task list
- [x] 1. Import drafts
  - [x] Cset version
  - [x] Expression items
  - [x] Members items #725
- [x] 2. Drafts finalized ? -> update TermHub
   a. (#398 should suffice?)
   b. Also ascertainable by when members are fetched (as in #725)

#### Implementation details
1. Drafts should appear normally everywhere like any other concept set.
2. Optional: Some additional indication ("draft" [badge/pill](https://getbootstrap.com/docs/4.0/components/badge/), icon, coloration, etc) that it is a draft
3. Optional: Toggles to show/hide drafts vs non-drafts

#### Problem 1. Only owner can download drafts
**Solution a:** Oauth
Then we should be able to fetch through Objects API.
([Slack discussion w/ Amin](https://cd2h.slack.com/archives/C047FEPM05U/p1692878673271559?thread_ts=1692305357.559459&cid=C047FEPM05U))

**Solution b:** Datasets API
Perhaps we can do a nightly download through the datasets API and import only drafts from there?
We could also maybe do some QC at that time and check that everything else is in the DB, essentially double checking that the objects API is working correctly.

### 319 drafts are currently present in TermHub
[319 cset draft versions in termhub.csv](https://github.com/jhu-bids/TermHub/files/12364173/319.cset.draft.versions.in.termhub.csv)

joeflack4:
Maybe should take some time to look at these two screenshots and understand what differences there are and why.

Most recent drafts on Enclave today:
![MicrosoftTeams-image](https://github.com/jhu-bids/TermHub/assets/13045020/d184fb82-5323-452b-b1ce-1bb77ab8a8e5)

Most recent drafts in TermHub today:
![MicrosoftTeams-image (1)](https://github.com/jhu-bids/TermHub/assets/13045020/a7e7adae-a83b-4ef3-935b-20bbe060d189)

joeflack4:
FYI, we got confirmation that drafts are visible in the API to us.
_[Maya wrote](https://cd2h.slack.com/archives/C047FEPM05U/p1696354494288849)_:
> currently for drafts of OMOPConceptSet, you'll be able to see from [UNITEConceptSetBulk@nih.gov](mailto:UNITEConceptSetBulk@nih.gov) if the source application = "TermHub"

Sigfried:
[...]

Issue #518, opened 2023-08-14
Bug: "Fetch failures - excessive members/items"


joeflack4:
## Overview
After a recent refresh, I checked the `fetch_audit` table and noticed cset `938394329` with message `Failed after 28000 members.`.

## Sub-task list
- [ ] 1. Script fails w/ HTTP error 406 ([log](https://github.com/jhu-bids/TermHub/actions/runs/5861503269/job/15891704377))
- [ ] 2. Need to successfully fetch `938394329` after '1' fixed.
- [ ] 3. Refresh fails to initiate GH action for "Fetch failures - excessive members/items"

Issue #499, opened 2023-07-31
DB refresh: fetch failures (excessive items): `KeyError`


joeflack4:
## Overview
There's a `KeyError` in the script for "fetch_failures_excess_items`.

I think the cause of it might be an LFS issue appearing again. For some reason even though it is downloading the dataset directly, somehow LFS came into play (I fixed this once) and when it tries to read it into `pandas`, it can't read the actual contents.

It's possible this isn't the reason, in which case it's probably some other dataframe column related issue.

## Additional notes
Even though this script is not running, this is probably not a huge priority since there is only 1 concept set where this happened in the last 6 months, and it's archived.

PR #725, opened 2024-03-09
Bugfixes: `fetch_failures_0_members()`

joeflack4:
## The crux
New drafts were sometimes never getting members fetched.

## Partially addresses
- #520

## Changes
    Bug fixes: fetch_failures_0_members()
    - Bug fix: Now does not ignore this issue when the cset is a draft.
    - Update: Now does not try to fix the issue mid-refresh because of new param try_fixing_issues_immediately=False. Added comments regarding rationale.
    - Update: Now addresses issus at end of refresh if for resolve_fetch_failures_0_members (set to True), but not for resolve_fetch_failures_excess_items (set to False), due to outstanding issues with that action.
    - Update: Fetch failures action / script: No longer runs for 2 hours. Just runs once and quits even if it could not fix the issue. Rationale: refresh runs every 20 minutes, so it will try again soon.
    - Update: Fetch audit table: add: error_datetime field
    - Bug fix: Action now doesn't fail if cset is still a draft when the refresh completes.

## Additional todo's if time
- [x] Add a timestamp for when the problematic cset w/ fetch failure first occurred (not just when it was successfully fetched)
- [x] Re-run refresh with `--since` to update csets with 0 members that are no longer drafts
- [x] FYI: to do later: `fetch_failures_excessive_items`: There are only cases of this once every year or two, so this is super low priority.

### DIFF
--- a/backend/db/initialize.py
+++ b/backend/db/initialize.py
@@ -59,6 +59,7 @@ DDL_FETCH_AUDIT = """
     "table" text not null,
     primary_key text not null,
     status_initially text not null,
+    error_datetime timestamp with time zone,
     success_datetime timestamp with time zone,
     comment text);"""
 
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -7,6 +7,9 @@ from datetime import datetime, timedelta
 from typing import Union
 import dateutil.parser as dp
 
+from backend.db.resolve_fetch_failures_0_members import resolve_failures_0_members_if_exist
+from backend.utils import call_github_action
+
 DB_DIR = os.path.dirname(os.path.realpath(__file__))
 BACKEND_DIR = os.path.join(DB_DIR, '..')
 PROJECT_ROOT = os.path.join(BACKEND_DIR, '..')
@@ -28,7 +31,8 @@ SINCE_ERR = '--since is more recent than the database\'s record of last refresh,
 # todo: What if 'since' is passed, but it is not the same date or before 'last_updated' in db? should print warning
 def refresh_db(
     since: Union[datetime, str] = None, use_local_db=False,  schema: str = CONFIG['schema'],
-    force_non_contiguity=False, buffer_hours=0
+    force_non_contiguity=False, buffer_hours=0, resolve_fetch_failures_0_members=True,
+    resolve_fetch_failures_excess_items=False
 ):
     """Refresh the database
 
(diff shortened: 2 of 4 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -485,10 +487,11 @@ def fetch_cset_and_member_objects(
                 insert_fetch_statuses([{'table': 'code_sets', 'primary_key': version_id, 'status_initially':
                     'fail-excessive-members', 'comment': f"Failed after {len(cset['member_items'])} members."}])
                 call_github_action('resolve-fetch-failures-excess-items')
-        if not cset['member_items'] and cset['expression_items'] and not cset['properties']['isDraft'] and handle_issues:
+        if not cset['member_items'] and cset['expression_items'] and handle_issues:
```
joeflack4, line 490: ## The bugfix
If it was a draft, we knew that members were not ready yet, so we didn't bother fetching. However, there was no way to then know that the draft was complete and get the members.

This quick and simple fix does not mark as a non-draft in TermHub, but it at least gets the members. So, it partially addresses #725.

I also included a note in `fetch_audit` about the cset being a draft at the time it wasn't able to fetch members.

backend/db/refresh.py
```diff
@@ -83,8 +92,14 @@ def refresh_db(
             counts_docs()
             raise err
         finally:
+            # Update status vars
             update_db_status_var('last_refresh_exited', current_datetime(), local)
             update_db_status_var('refresh_status', 'inactive', local)
+            # Routine: Check for and resolve any open fetch failures
+            if resolve_fetch_failures_excess_items:
+                call_github_action('resolve-fetch-failures-excess-items')
+            if resolve_fetch_failures_0_members:
```
joeflack4, line 101: `resolve_fetch_failures_0_members` is on.

But `resolve_fetch_failures_excess_items` is off due to outstanding issues #499 and #518. but it hardly matters, as it almost never happens.

enclave_wrangler/objects_api.py
```diff
@@ -410,15 +410,20 @@ def find_and_add_missing_csets_to_db(con: Connection = None, schema=SCHEMA):
 
 
 def fetch_cset_and_member_objects(
-    since: Union[datetime, str] = '', codeset_ids: List[int] = [], handle_issues=True, verbose=False
+    since: Union[datetime, str] = '', codeset_ids: List[int] = [], flag_issues=True, verbose=False,
+    try_fixing_issues_immediately=False
```
joeflack4, line 414: ## `try_fixing_issues_immediately=False`
Now, by default, no longer tries to fix mid-refresh; fixes at end of refresh, as it takes members time to appear.

backend/db/initialize.py
```diff
@@ -59,6 +59,7 @@
     "table" text not null,
     primary_key text not null,
     status_initially text not null,
+    error_datetime timestamp with time zone,
```
joeflack4, line 62: ## Add: `error_datetime`
Added field. Not using in codebase currently; just there for troubleshooting purposes / data collection / possible future use. Added via:
```sql
ALTER TABLE your_table_name
ADD COLUMN error_datetime TIMESTAMP WITH TIME ZONE DEFAULT current_timestamp;
```

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -35,71 +36,104 @@ def _report_success(
...
-            _report_success(success_cases, failure_lookup, 'Success result: Found members', use_local_db)
-
-        # Sleep
+        # Sleep or exit
+        if not loop:
+            break
         time.sleep(polling_interval_seconds)
 
+    # Remove drafts from failures (as csets don't have members until drafts are finalized)
+    failed_cset_ids = [cset_id for cset_id in failed_cset_ids if cset_is_draft_map[cset_id]]
```
joeflack4, line 124: ## Don't error on drafts
Action now doesn't fail if cset is still a draft when the refresh completes.

PR #729, opened 2024-03-13
Mitigate proliferation of idle connections: SqlAlchemy Con --> Engine

joeflack4:
addresses #740 
though for some reason I can't link it on the right sidebar

## Changes
Connection to Engine
- Update: Usage of SqlAlchemy Engine instead of Connection object in order to prevent proliferation of idle connections.

### DIFF
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -206,6 +206,41 @@ def refresh_derived_tables(
             break
 
 
+ENGINES = {
+    # 'n3c': get_engine(),
+    # 'public': get_engine(schema='public')
+}
+
+
+def get_engine(isolation_level='AUTOCOMMIT', schema: str = SCHEMA, local=False):
+    """Connect to db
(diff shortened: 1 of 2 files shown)


backend/db/utils.py
```diff
@@ -206,6 +206,41 @@ def refresh_derived_tables(
             break
 
 
+ENGINES = {
+    # 'n3c': get_engine(),
+    # 'public': get_engine(schema='public')
+}
+
+
+def get_engine(isolation_level='AUTOCOMMIT', schema: str = SCHEMA, local=False):
```
joeflack4, line 215: ## `get_engine()`
@Sigfried Gonna just leave the work we did in this open PR. This is a low priority issue now that we've resolved the phone call in the house and also got an action that closes extra idle connections.

joeflack4:
Closing for cleanup but the issue remains open, and can be addressed by this.

PR #731, opened 2024-03-13
AMIA paper: Usage statistics, etc

joeflack4:
## Changes
AMIA paper 2024: stats
- Add: stats.py: For producing tables and figures related to usage and possibly other statistics.

General
- Add: /misc/: For things that are not used very often, and not part of the core application.
- Moved: Several root folders into /misc/


misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,33 @@
+"""AMIA paper 2024: Statistics
```
joeflack4, line 1: ## General updates
1. Add: `/misc/`, and moved several folders into there.
2. Update: `termhub-vocab`

I'm not sure how (2) happened, but for right now I'll permit it.

@Sigfried (1) is just something that's bugged me for a while. I needed somewhere to put this and I didn't want a proliferation of stuff in the root. If you want though, I can do all of this moving of stuff in a separate PR.

backend/routes/db.py
```diff
@@ -725,13 +726,15 @@ def next_api_call_group_id() -> int:
...
-def usage_query():
+def usage_query(verbose=True):
     """Query for usage data"""
+    t0 = datetime.now()
     with get_db_connection() as con:
         data = sql_query(con, """
             SELECT DISTINCT r.*, array_sort(g.api_calls) api_calls, g.duration_seconds, g.group_start_time
             FROM public.apiruns_grouped g
             RIGHT JOIN public.api_runs r ON g.api_call_group_id = r.api_call_group_id""")
+    print(f'usage_query(): Fetched {len(data)} records in n seconds: {(datetime.now() - t0).seconds}')
```
joeflack4, line 737: ## `usage_query()` refactor
I added this function to `develop` already. Here, I added a printout of how long it takes. I haven't run since I added this line, but earlier when we were on call it felt like it completed in 3 minutes or so.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,60 @@
...
+        data: List[Dict] = usage_query()
+        df = pd.DataFrame(data)
+        # Save
+        t0 = datetime.now()
+        df.to_csv(CSV_PATH, index=False)
+        print(f'Wrote {len(df)} records to CSV in n seconds: {(datetime.now() - t0).seconds}')
+    else:
+        # Read
+        t0 = datetime.now()
+        df = pd.read_csv(CSV_PATH)
```
joeflack4, line 46: ## CSV
Reads CSV if it exists in local (git ignored) dir. Else, it fetches data and writes the CSV. FYI on my PC it takes 7 seconds to read the CSV. Also the CSV is a similar size to the JSON; almost 3GB.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,73 @@
...
+        print(f'Read {len(df)} records from CSV in n seconds: {(datetime.now() - t0).seconds}')
+    return df
+
+
+def preprocess(df: pd.DataFrame, verbose=True) -> pd.DataFrame:
```
joeflack4, line 51: ## Dev IP preprocessing
Maybe I should've added this to `usage_query()`, but I added it to my `amia_paper_2024/stats.py`.

I'm surprised that it's showing this few records for you, and 0 records for me:
> ```
> Filtered out n records based on IP addresses of each dev:
>   - Siggie: 62
>   - Joe: 0
> ```

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,77 @@
...
+    developer_ips = {'Siggie': '216.164.48.98', 'Joe': '174.99.54.40'}
+    df2 = df[~df['client'].isin(list(developer_ips.values()))]
+    if verbose:
+        print('Filtered out n records based on IP addresses of each dev:')
+        for dev, ip in developer_ips.items():
+            df_i = df[df['client'] == ip]
+            print(f'  - {dev}: {len(df_i)}')
+    # Remove records with no 'API call group'
+    # - These are cases from before we added this feature, or where we called the backend directly.
+    df2 = df2[~df2['api_call_group_id'].isna()]
```
joeflack4, line 63: ## Null `api_call_group_id` preprocessing
I removed these. I didn't count how many rows, but there weren't many. Like hundreds out of the total ~52k.

misc/amia_paper_2024/stats.py
joeflack4: Also, surprisingly, all of these entries are cases where `api_call_group_id = NULL`. If I remove all of those cases, it doesn't show any records for dev IP addresses.

joeflack4: FYI I added more ideas etc to a [comment in the paper](https://livejohnshopkins-my.sharepoint.com/:w:/r/personal/sgold15_jh_edu/Documents/JBI-paper-2022-revision/[codesethub-amia-2023-12-draft.docx](https://livejohnshopkins-my.sharepoint.com/:w:/r/personal/sgold15_jh_edu/Documents/JBI-paper-2022-revision/codesethub-amia-2023-12-draft.docx?d=w00c7ee766f6c4f4a8057e03d88d1eaa8&csf=1&web=1&e=oWRpIB&nav=eyJjIjoxNzg1MjgyNTV9)?d=w00c7ee766f6c4f4a8057e03d88d1eaa8&csf=1&web=1&e=oWRpIB&nav=eyJjIjoxNzg1MjgyNTV9).

backend/routes/db.py
```diff
@@ -725,23 +726,100 @@ def next_api_call_group_id() -> int:
...
+        #     SELECT DISTINCT r.*, array_sort(g.api_calls) api_calls, g.duration_seconds, g.group_start_time
+        #     FROM public.apiruns_grouped g
+        #     RIGHT JOIN public.api_runs r ON g.api_call_group_id = r.api_call_group_id""")  # 530,232
+        data: List[RowMapping] = sql_query(con, """
             SELECT DISTINCT r.*, array_sort(g.api_calls) api_calls, g.duration_seconds, g.group_start_time
-            FROM public.apiruns_grouped g
-            RIGHT JOIN public.api_runs r ON g.api_call_group_id = r.api_call_group_id""")
+            FROM public.api_runs r
+            LEFT JOIN public.apiruns_grouped g ON g.api_call_group_id = r.api_call_group_id
+            WHERE g.api_call_group_id != -1 AND g.api_call_group_id IS NOT NULL;""")  # 13,210
```
joeflack4, line 742: ## Fixing results (530,232 --> 13,210)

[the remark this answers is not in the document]
joeflack4: I definitely want to understand why/if that's the case. The only reason I can think of is that left/inner will effectively be the same now because now we expect that the distinct values for `api_call_group_id` are the same in both tables. But that seems like what we want.

Certainly for some of the statistics, I'll want to get rid of these NULL and -1 groups. But perhaps for others, I think now that you might want me to sample from all of the `api_runs`, regardless, possibly. E.g. total number of errors. But any cases where I do that, now our stats will be proportionally out of sync.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,160 @@
...
+            if verbose:
+                print(f'Wrote {len(df)} records to {path} in n seconds: {(datetime.now() - t0).seconds}')
+        else:
+            # Read
+            t0 = datetime.now()
+            df = pd.read_csv(path)
+            if verbose:
+                print(f'Read {len(df)} records from {path} in n seconds: {(datetime.now() - t0).seconds}')
+    # Mods
+    df['client_ip'] = df['client'].apply(lambda x: x.split(':')[0])
```
joeflack4, line 77: ## IP address stuff
The tot IP addresses without dev IPs is 234, and with dev IPs is 230. @Sigfried You may wonder why the difference is >2. It's because of localhost:

The 4 'dev ips':
- '174.99.54.40'
- 'no forward; 127.0.0.1 (no loc, err 106)'
- '127.0.0.1'
- '216.164.48.98'

<details><summary>Additional details</summary>
<p>

Code I used to get this diff:
```py
  # developer_ips = {'Siggie': '216.164.48.98', 'Joe': '174.99.54.40'}
  ip = set(df_apiruns['client_ip'].unique())
  ip_nodev = set(preprocess_dev_ips(df_apiruns, verbose)['client_ip'].unique())
  diff = ip.difference(ip_nodev)
```

Note that `client_ip` is a new column. This is the result of the operation:
`df['client_ip'] = df['client'].apply(lambda x: x.split(':')[0])`

If I did the diff on `client`, it would be:
- 'no forward; 127.0.0.1 (no loc, err 106)'
- '216.164.48.98:55408; 169.254.130.3'
- '127.0.0.1'
- '174.99.54.40: Raleigh, no region'
- '216.164.48.98:61752; 169.254.130.6'
- '216.164.48.98: Bethesda, Maryland'
- '174.99.54.40: Raleigh, North Carolina'
- '216.164.48.98:57532; 169.254.130.6'
- '216.164.48.98: Takoma Park, Maryland'
- '216.164.48.98: Lake Ridge, Virginia'
- '216.164.48.98:55659; 169.254.130.6'
- '216.164.48.98: Wheaton, Maryland'
- '216.164.48.98: Takoma Park, no region'

</p>
</details>

misc/amia_paper_2024/stats.py
joeflack4: Now there are none of them after my new query. It seems these only exist in the`apiruns_grouped` tablee

misc/amia_paper_2024/stats.py
joeflack4: I moved (1) to another PR.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,183 @@
...
+    summary['IP addresses'] = len(df_apiruns['client_ip'].unique())
+    summary['Errors'] = df_w_groups_filtered[df_w_groups_filtered['result'].str.lower().str.contains('error')][
+        'api_call_group_id'].nunique()
+
+    # Value set combos
+    df_get_csets = df_w_groups_filtered[df_w_groups_filtered['api_call'] == 'get-csets']
+    summary['Value set combos'] = len(set([tuple(x) for x in df_get_csets['codeset_ids']]))
+
+    # todo: temp: analysis
+    #  - analyze uniqueness of get-csets calls within an API call group session
```
joeflack4, line 137: ## Value set combos
**How I'm ascertaining (for now)**: By unique calls on `get-csets`.
I did it this way in the interest of time. It is not a bad proxy for the truth., and may even be revealed to be the truth, after addressing data quality issues.

**Data quality & strength of this ascertainment method**: I did not expect this, but out of 2,944 `get-csets` calls, 30 `api_call_group_id` sessions where the same call (same combo of csets) was being called multiple times. I don't know if this is a programming error or an artefact of lag, or a gap in my understanding. There was only 1 case in which there were 2 `get-csets` calls in a single group which had different csets. All other sessions had only 1 call per session.

<p>

See comments after variable names for values.

```py
    # Value set combos
    df_get_csets = df_w_groups_filtered[df_w_groups_filtered['api_call'] == 'get-csets']
    summary['Value set combos'] = len(set([tuple(x) for x in df_get_csets['codeset_ids']]))

    # todo: temp: analysis
    #  - analyze uniqueness of get-csets calls within an API call group session
    # diff_get_csets_vs_groups=35; i feel like this should be equal to some calc involving next 4 vars, but not sure
    # noinspection PyUnusedLocal
    diff_get_csets_vs_groups = len(df_get_csets) - len(df_get_csets['api_call_group_id'].unique())  # 35
    diff_codeset_ids_in_group__n_instances = 0  # 1
    diff_codeset_ids_in_group__n_calls = 0  # 2
    mult_get_cset_in_session__n_instances = 0  # 30
    mult_get_cset_in_session__n_calls = 0  # 64
    # todo: maybe convert to str in pre-processing instead, if need 'str' more than temporarily
    df_get_csets['codeset_ids'] = df_get_csets['codeset_ids'].astype(str)
    for group_id, group_data in df_get_csets.groupby('api_call_group_id'):
        if len(group_data) > 1 and len(group_data['codeset_ids'].unique()) > 1:
            diff_codeset_ids_in_group__n_instances += 1
            diff_codeset_ids_in_group__n_calls += len(group_data)
        elif len(group_data) > 1:
            mult_get_cset_in_session__n_instances += 1
            mult_get_cset_in_session__n_calls += len(group_data)
```

</p>
</details> 

**Why I chose this method:** I didn't have knowledge of what the unique ways in which an `api_call_group_id` session could be started. As far as I could see, they almost always started off with `get-csets`, and in cases where they did not, they were soon followed by a call to `get-csets`, almost always calling the sam set of IDs.

**Alternate solutions**: I suppose I could look at the first timestamp for each session, but I'm not sure it is always representative of the cset combos being primarily queried.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,216 @@
...
+    # Stats: With def IPs filtered out
+    df_out_dev0: pd.DataFrame = summary_stats(df_apiruns_dev0, df_w_groups_filtered_dev0)
+
+    # Join different output datasets
+    df_out = df_out_dev1.merge(df_out_dev0.rename(columns={'Value': 'Value_no_dev'}), on='Measure', how='outer')
+    df_out.to_csv(OUT_CSV_PATH_ALL, index=False)
+
+    # Plots ---
+    # todo: consider dev + non-dev in same plot
+    # Histogram: Value set n selections
```
joeflack4, line 194: ## Histogram: Value set n selections

code_set_comparison_size


code_set_comparison_size_-_dev_data_included


I was surprised to see that the distribution was very similar. Although I'm not comparing dev to non-dev. Just dev+non-dev to non-dev. Looking at dev alone would probably be different.

Also, if we check w/out filtering the -1 `api_call_group_id`, the max comparison is bigger:
```py
df_apiruns['len_codeset_ids'] = df_apiruns['codeset_ids'].apply(lambda x: len(x))
print(max(df_apiruns['len_codeset_ids']))
>>> 20
```

I think these numbers are small enough that scaling / log scaling is not necessary.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,236 @@
...
+):
+    """Generate summary statistics for given datasets
+
+    :param df_apiruns: Includes full logs, even for groups that have erroneous -1 as the api_call_group_id. This is only
+     used for tallying unique IP addresses.
+    :param df_w_groups_filtered: This is a JOIN of apiruns and apiruns_grouped tables, with erroneous api_call_group_id
+     of -1 filtered out."""
+    # Get summary statistics
+    summary = {}
+    summary['Total log records'] = len(df_apiruns)
```
joeflack4, line 139: Ok I see what we're doing now

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,236 @@
...
+    :param use_cache: If True, will use most recent local CSV instead of calling the database"""
+    # Initial setup ---
+    setup()
+    df_apiruns: pd.DataFrame = get_dataset_with_mods(api_runs_query, USAGE_UNJOINED_CSV_PATH, use_cache, verbose)
+    df_w_groups_filtered: pd.DataFrame = get_dataset_with_mods(usage_query, USAGE_JOINED_CSV_PATH, use_cache, verbose)
+    df_apiruns_dev0: pd.DataFrame = preprocess_dev_ips(df_apiruns, verbose)
+    df_w_groups_filtered_dev0: pd.DataFrame = preprocess_dev_ips(df_w_groups_filtered, verbose)
+
+    # Table ---
+    # Stats: With dev IPs included
```
joeflack4, line 188: thanks for typo fix

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,236 @@
...
+        title_affix = f' - {name_suffix}' if name_suffix else ''
+        title = f'Code set comparison size{title_affix}'
+        filename = title.replace(' ', '_').lower()
+        plt.title(title)
+        if verbose:
+            plt.show()
+        plt.savefig(OUTDIR / f'{filename}.png')
+        plt.clf()
+
+    for df, name_suffix in ((df_w_groups_filtered, 'Dev data included'), (df_w_groups_filtered_dev0, '')):
```
joeflack4, line 218: Haven't looked at this new plot but will!

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,258 @@
...
+            print(f'  - {dev}: {len(df_i)}')
+    df2 = df[~df['client_ip'].isin(list(developer_ips.values()))]
+    # Devs running locally
+    count_before = len(df2)
+    df2 = df2[df2['host'].isin(['dev', 'prod'])]
+    if verbose:
+        print('Filtered out n records based devs running locally: ',  count_before - len(df2))
+    # Test cases being run by GitHub actions: IPs unknown
+    test_cases = [
+        '[1000002363, 1000002657, 1000007602, 1000013397, 1000010688, 1000015307, 1000031299]',  # many small
```
joeflack4, line 117: ## Filter: automated test logs
@Sigfried I am filtering these out now. This was about 33% of the data.

I did a face check and confirmed as well by looking at the timestamps. All of these removals I saw were around 3AM our time, which is what I'd expect from the cronjob:
> `- cron: '0 8 * * *'  # every day 8am GMT (3-4am EST/EDT)`

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,417 @@
...
+        """Relabel column vals: log10"""
+        min_val = val.split(' - ')[0]
+        n = min_val.count('0')
+        return f'10^{n}-10^{n+1}'
+    df['concepts_in_call'] = df['concepts_in_call'].apply(relabel)
+    return df
+
+
+# todo: y axis also in log scale
+def plot_concepts_in_calls():
```
joeflack4, line 134: ## Plot: Concepts in calls
Here's what I got.

For the small version, I wasn't sure the best way to do the xlabels and still get them to fit in a small space. I've included 2 variations.

I also realize that the y data is too small in some bins to be shown. This can be fixed with some more logarithmic scaling, but I'll move onto other plots and I can come back to this if we have time.

n_concepts_in_api_calls - small

n_concepts_in_api_calls - small2

n_concepts_in_api_calls

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,461 @@
...
+        # for df, name_suffix in ((df_apiruns, 'Dev data included'), (df_apiruns_dev0, '')):
+        data: List = list(df_i['week'])
+        bins = len(set(data))
+        # Render
+        plot_render(data, title, 'Week', 'API calls', bins, small, is_timeseries=True)
+
+    # Histogram: Session duration (duration_seconds)
+    for df_i, name_suffix in datasets:
+        # Title
+        title = f'Duration, session API calls{f" - {name_suffix}" if name_suffix else ""}'
```
joeflack4, line 398: ## Duration, session API calls
There's a whole lot of variation in the <0 range, but this is good enough I think. Adding more bins could make us see that better though.

duration_session_api_calls_log_scale - small

<details><summary>Big version</summary>
<p>

duration_session_api_calls_log_scale

</p>
</details> 

### `duration_seconds` vs `process_seconds`
@Sigfried I was a little unsure of what to use here. I think you intended for me to use `duration_seconds`. It was consistent with what I was able to calculate from the timestamps in each API call group. However, some things I wonder:

1. Was it indeed better that I chose `duration_seconds`?

2. What is `process_seconds`, and why is it different than `duration_seconds`?
Does it have something to do with async? Like are multiple things being processed simultaneously so that the sum time of each call doesn't equate with how long it takes for the user to get all the data?

3. Is there an issue with `duration_seconds` calculation from timestamps?
I'm concerned there might be a problem with this calculation. I think it would be better for each API call to actually have 2 timestamps: start and end time. Then, we could consider that the duration of an API call group would be the (end time of last call) - (start time of the 1st call). However, given that each call only has 1 timestamp, we can't calculate this, and I am assuming each timestamp represents the start time (or end time) only. If that's the case, then I think the duration of the final call is not being included in `duration_seconds`.

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,490 @@
...
+    # data = [round(x, 1) for x in data]  # lower decimal precision didn't make any diff
+    # - int or deca seconds also didn't make a difference
+    bins = 50
+    # Render
+    plot_render(
+        data, title, 'n seconds', 'n sessions', bins, small, is_timeseries=False, log_scale=True,
+        xtick_step=10)
+
+    # todo: alternate view where I have n concepts on y axis, and n seconds on x
+    # Histogram: Session duration divided by concept count
```
joeflack4, line 435: ## Plot: Session duration divided by concept count

misc/amia_paper_2024/stats.py
```diff
@@ -0,0 +1,512 @@
...
+    df_i2 = df_i2[~df_i2['duration_sec_per_concept'].isin([np.inf, -np.inf, ''])]  # idk how '' snuck in
+    data: List[float] = df_i2['duration_sec_per_concept'].tolist()
+    # noinspection PyTypeChecker
+    data = [x for x in data if x]  # filter null ''s
+    bins = 30
+    # Render
+    plot_render(
+        data, title, 'n seconds / concept', 'instances', bins, small, log_scale=True)
+
+    # Histogram: User queries
```
joeflack4, line 453: ## Plot: User queries

### N queries made by users
| No binning | Binning |
|--------|--------|
| n_queries_made_by_users_-_no_bins_log_scale - small | n_queries_made_by_users_log_scale - small | 

<details><summary>Big versions</summary>
<p>

n_queries_made_by_users_-_no_bins_log_scale

n_queries_made_by_users_log_scale

</p>
</details>

backend/db/ddl-18-apirun_groups.jinja.sql
```diff
@@ -11,40 +10,89 @@ CREATE TABLE IF NOT EXISTS public.api_runs (
...
-        WHERE api_call_group_id IS NOT NULL
-    )
+    ),
+groups_broke_by_long_gap AS (
+  SELECT *,
+    host || '-' || ip3 || '-' || COALESCE(api_call_group_id::text, 'no_call_grp') || '-' ||
+        SUM(new_group_flag) OVER (PARTITION BY host, ip3, api_call_group_id
+            ORDER BY timestamp::timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS group_id
+  FROM RankedGroups
+  WHERE NOT ((week IN ('2024-03-04', '2024-03-11') AND ip3='130.44.174') -- tons of these, don't know why
```
joeflack4, line 54: Strange

PR #747, opened 2024-04-07
`initialize()` & DDL updates

joeflack4:
## Changes
- Update: DDL for concepts_with_counts
  - Delete: commented out code dropping concepts_with_counts_ungrouped

### DIFF
--- a/backend/db/ddl-14-concepts_with_counts.jinja.sql
+++ b/backend/db/ddl-14-concepts_with_counts.jinja.sql
@@ -20,5 +20,5 @@ CREATE TABLE IF NOT EXISTS {{schema}}concepts_with_counts{{optional_suffix}} AS 
 
 CREATE INDEX cc_idx1{{optional_index_suffix}} ON {{schema}}concepts_with_counts{{optional_suffix}}(concept_id);
 
--- the following drop table is causing errors with the initialize script
--- DROP TABLE {{schema}}concepts_with_counts_ungrouped{{optional_suffix}} CASCADE;
+-- Index cwcgin takes some time to complete
+CREATE INDEX cwcgin{{optional_index_suffix}} ON concepts_with_counts{{optional_suffix}} USING gin (concept_name gin_trgm_ops);
--- a/backend/db/initialize.py
+++ b/backend/db/initialize.py
@@ -3,6 +3,9 @@
 1. Initialize the database if it doesn't exist
 2. Download CSV files from the enclave
 3. Load the data from the CSV files into the database
+
+todo's (minor)
+  1. Consider moving DDL strings here into separate DDL files in backend/db/
 """
 import os
 import sys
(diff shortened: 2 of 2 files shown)


backend/db/ddl-14-concepts_with_counts.jinja.sql
```diff
@@ -19,6 +19,3 @@ CREATE TABLE IF NOT EXISTS {{schema}}concepts_with_counts{{optional_suffix}} AS
     ORDER BY concept_id, domain );
 
 CREATE INDEX cc_idx1{{optional_index_suffix}} ON {{schema}}concepts_with_counts{{optional_suffix}}(concept_id);
-
--- the following drop table is causing errors with the initialize script
--- DROP TABLE {{schema}}concepts_with_counts_ungrouped{{optional_suffix}} CASCADE;
```
joeflack4, line 24: ## Removed comment
We have diff DDLs for `concepts_with_counts_ungrouped` and `concepts_with_counts`, so I don't think that dropping the former is something we want to do. It's currently a table that we have, and I think we're using. If we ever do want to drop it during creation of `concepts_with_counts`, then we should make a single DDL file for the both of them.

backend/db/initialize.py
```diff
@@ -49,24 +54,39 @@
...
+# DDL: PostgreSQL extensions
+#  Extension name | Description                   | Utilized as of 2024/04/07
+#  PG_TRGM        | Trigrams                      | Yes
+#  BTREE_GIN      | Generalized Inverted Index    | No
+#  BTREE_GIST     | Generalized Search Tree Index | No
+# Enabling extensions requires two steps:
+# 1. Enable them in the cloud, e.g. Azure: https://portal.azure.com/#@live.johnshopkins.edu/resource/subscriptions/fe24df19-d251-4821-9a6f-f037c93d7e47/resourceGroups/JH-POSTGRES-RG/providers/Microsoft.DBforPostgreSQL/flexibleServers/termhub/serverParameters
+# 2. Run SQL: CREATE EXTENSION <EXTENSION>;
+# If, when running initialize() there is an error because can't create the extension, try step (1) and enable in Azure.
+DDL_EXTENSIONS = """
```
joeflack4, line 86: Extensions

backend/db/ddl-14-concepts_with_counts.jinja.sql
```diff
@@ -20,5 +20,5 @@ CREATE TABLE IF NOT EXISTS {{schema}}concepts_with_counts{{optional_suffix}} AS
 
 CREATE INDEX cc_idx1{{optional_index_suffix}} ON {{schema}}concepts_with_counts{{optional_suffix}}(concept_id);
 
--- the following drop table is causing errors with the initialize script
--- DROP TABLE {{schema}}concepts_with_counts_ungrouped{{optional_suffix}} CASCADE;
\ No newline at end of file
+-- Index cwcgin takes some time to complete
+CREATE INDEX cwcgin{{optional_index_suffix}} ON concepts_with_counts{{optional_suffix}} USING gin (concept_name gin_trgm_ops);
```
joeflack4, line 24: New index

PR #748, opened 2024-04-08
Update `make test-missing-csets`

joeflack4:
## Changes
- Update: makefile goal: test-missing-csets. Now this test fails only if csets are older than 30 minutes.

### DIFF
--- a/enclave_wrangler/objects_api.py
+++ b/enclave_wrangler/objects_api.py
@@ -12,21 +12,19 @@ TODO's
  concept_set_version_item_rv_edited_mapped.
  3. All _db funcs / funcs that act on the DB (and unit tests) should be in backend/, not enclave_wrangler.
 """
+import dateutil.parser as dp
 import json
 import os
 import sys
 from argparse import ArgumentParser
-
-from datetime import datetime
+from datetime import datetime, timedelta
 from pathlib import Path
 from typing import List, Dict, Set, Tuple, Union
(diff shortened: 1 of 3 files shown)


enclave_wrangler/objects_api.py
```diff
@@ -358,13 +358,34 @@ def get_bidirectional_csets_sets(con: Connection = None) -> Tuple[Set[int], Set[
     db_codeset_ids: Set[int] = set(sql_query_single_col(con, 'SELECT codeset_id FROM code_sets'))
     # Set 2 of 2: In the enclave
     enclave_codesets = fetch_all_csets()
-    # Set difference & return
     enclave_codeset_ids: Set[int] = set([cset['codesetId'] for cset in enclave_codesets])
+
     return db_codeset_ids, enclave_codeset_ids
 
 
+def find_missing_csets_within_threshold(age_minutes=30, con: Connection = None) -> Set[int]:
```
joeflack4, line 366: ## Refresh test now fails if missing tests >30 minutes old
Since the refresh runs very often, we're getting failures surprisingly often, where, during the course / finishing of the DB refresh, a brand new cset was added, and when it got to the test at the end, it would fail. And then the next refresh would succeed, because the cset would be picked up.

Since these run every 20 minutes, I picked 30 minutes as a threshold. Now the action will only fail if a cset is detected that is older than that.

Issue #637, opened 2023-12-04
Need server timeouts


joeflack4:
## Overview
Need long-running backend tasks to time out eventually. According to[ this](https://docs.gunicorn.org/en/latest/settings.html#timeout), gunicorn is supposed to have a default timeout of 30 seconds; but I don’t see a way to set a timeout for uvicorn. I found some [instructions for timeout middleware](https://github.com/tiangolo/fastapi/issues/1752#issuecomment-682579845) but it’s not clear from the discussion there how well it works.

joeflack4:
Is the second one here helpful?:
https://www.uvicorn.org/settings/#timeouts
> `--timeout-keep-alive <int>` - Close Keep-Alive connections if no new data is received within this timeout. Default: 5.
> `--timeout-graceful-shutdown <int>` - Maximum number of seconds to wait for graceful shutdown. After this timeout, the server will start terminating requests.

joeflack4:
I couldn’t find either on Google or through ChatGPT. Surprised. I even tested both commands they have there in docs and didn’t work.

Issue #601, opened 2023-11-01
Add concepts to csets


joeflack4:
Related:
- #600

joeflack4:
When we search, should show the concepts most popular at the top. Not alphabetical.

joeflack4:
I think some more of these check boxes have been completed, but not sure which.

PR #749, opened 2024-04-08
MVP: Add concepts to csets - string search

Sigfried:
- addresses #774
  - addresses #769
- addresses #781
- addresses #784

May have addressed:
- #787


backend/routes/db.py
```diff
@@ -199,6 +199,96 @@ def get_all_csets(con: Connection = None) -> Union[Dict, List]:
...
 
+
+@router.get("/next-api-call-group-id")
+def next_api_call_group_id() -> int:
+    """Get next API call group ID"""
+    with get_db_connection() as con:
+        id = sql_query_single_col(con, "SELECT nextval('api_call_group_id_seq')")[0]
+    return id
+
+
```
Sigfried, line 210: Adding new routes:
 - start-session
 - continue-session

backend/db/utils.py
```diff
@@ -50,6 +51,24 @@
 SCHEMA = CONFIG["schema"]
```
Sigfried, line 53: New route:
 - log-concepts

backend/db/ddl-21-sessions.jinja.sql
```diff
@@ -0,0 +1,24 @@
...
+    UNIQUE(session_id, concept_id)
+);
+/* use like this:
+        INSERT INTO session_concept (session_id, concept_id)
+        VALUES (:session_id, :concept_ids)
+        ON CONFLICT (session_id, concept_id) DO NOTHING
+ */
+
+CREATE INDEX session_conceptidx{{optional_index_suffix}} ON {{schema}}session_concept{{optional_suffix}}(session_id, concept_id);
+CREATE INDEX session_conceptidx2{{optional_index_suffix}} ON {{schema}}session_concept{{optional_suffix}}(session_id, concept_id);
```
joeflack4, line 24: @Sigfried This `session_conceptidx2` is a duplicate. Did you mean to add a second index?

backend/db/ddl-21-sessions.jinja.sql
```diff
@@ -0,0 +1,24 @@
+DROP TABLE IF EXISTS public.sessions{{optional_suffix}} CASCADE;
```
joeflack4, line 1: ## Replaced `{{schema}}` with `public.`
Everywhere in this file.

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+sessions_cache: Dict = {}
+session_stale_hours = 24
+
+
+@router.get("/start-session")
+def start_session():
+    return get_session(None)
+
+
+def get_session(session_id: int):
```
joeflack4, line 287:
```suggestion
def get_session(session_id: Union[int, None]):
```

or

```suggestion
def get_session(session_id: Union = None):
```
and change `return get_session(None)` to `return get_session()`

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+    if session_id:
+        if session_id not in sessions_cache:
+            # raise HTTPException(status_code=404, detail="Session not found")
+            raise Exception(f"Session {session_id} not found")
+        return sessions_cache[session_id]
+    session = {
+        "session_id": next_api_call_group_id(),
```
joeflack4, line 294: I see--so sessions and API groups share the same ID range/ mechanism?

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+def update_session(session_id: int, concepts: List[Dict]):
+    session = get_session(session_id)
+    session["concept_ids"].extend([c["concept_id"] for c in concepts])
+    session["last_used"] = datetime.now()
+    sessions_cache[session_id] = session
+    clear_old_sessions()
+    return session
+
+
+@throttle(wait = 600) # 10 minutes
```
joeflack4, line 311: Oh interesting, what does this do? I've never seen.

Does this come from this package?: https://pypi.org/project/throttle/
If so, we should:
- [ ] Update `requirements.txt` and `requirements-unlocked.txt`

_edit_: I see now you made your own throttle decorator. If you're having any issues with it, or if we just prefer to use an external package, we could do so, but I'm fine with either.

joeflack4, line 311:
```suggestion
@throttle(wait = 60 * 10) # 10 minutes
```

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
     return await get_concepts_route(request, id=id, table=table)
 
 
+sessions_cache: Dict = {}
+session_stale_hours = 24
```
joeflack4, line 279:
```suggestion
SESSION_STALE_HOURS = 24
```

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
     return await get_concepts_route(request, id=id, table=table)
 
 
+sessions_cache: Dict = {}
```
joeflack4, line 278:
```suggestion
SESSIONS_CACHE: Dict = {}
```
IDK if this is plural by accident (other variables seem to be singular), but at the moment I have no objections.

backend/routes/db.py
joeflack4: So the idea is that we limit session clearing to 1x every 10 minutes, and when we trigger a clear, we clear anything older than 24 hours?

joeflack4: I assume the throttling is to save on performance, but does it help that much? Is this function really going to get triggered that often and, if so, is it really that taxing to clear attempt to clear the cache?

backend/routes/db.py
joeflack4, line 279: ## Cache clearing: time-based vs LRU
I thought we were going to have a fixed amount of search results cached on the server? Are we not going to do that? I don't strongly object.

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+    session = {
+        "session_id": next_api_call_group_id(),
+        "session_start": datetime.now(),
+        "concept_ids_seen": [],
+        "searches": {},
+    }
+    return session
+
+
+def update_session(session_id: int, concepts: List[Dict]):
```
joeflack4, line 302: ## `update_session()`: 0 usages?
Is this just temporary? I saw you commented out old method that was using this func. Will this func be used via some other method?

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+    #   for desc, prefix with -
+    sort_strs = sort_by.split('|')
+    sort_cols = []
+    for s in sort_strs:
+        desc = False
+        if s[0] == '-':
+            desc = True
+            s = s[1:]
+        if s not in sort_columns:
+            raise f"invalid sort_by: {s}"
```
joeflack4, line 331:
```suggestion
            raise ValueError(f"invalid sort_by: {s}")
```

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+        if datetime.now() - sessions_cache[session_id]["last_used"] > timedelta(hours=session_stale_hours):
+            del sessions_cache[session_id]
+    pdump(sessions_cache)
+
+
 @router.get("/concept-search")
-async def get_concepts_route(search: str) -> List:
-    q = """
-      SELECT *
+async def _concept_search(search_str: str, sort_by: str = "-total_cnt|vocabulary_id|concept_name") -> List[int]:
```
joeflack4, line 320: Looks good

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+    sort_columns = {"total_cnt", "concept_name", "vocabulary_id", "domain_id", "concept_class_id"}
```
joeflack4, line 321:
```suggestion
    valid_sort_columns = {"total_cnt", "concept_name", "vocabulary_id", "domain_id", "concept_class_id"}
```
To better disambiguate the purpose of `sort_cols` from `sort_columns`.

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+            s = s[1:]
+        if s not in sort_columns:
+            raise f"invalid sort_by: {s}"
+        s = f"{s} DESC" if desc else s
+        sort_cols.append(s)
+    q = f"""
+      SELECT concept_id
       FROM concepts_with_counts
-      WHERE concept_name ILIKE %s
+      WHERE concept_name ILIKE :search_str
```
joeflack4, line 337: ## Tri-grams?
I see `ILIKE`, but I don't see anything about that tri-gram extension here. Are we ditching that? If so, why?

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+def next_api_call_group_id() -> int:
+    """Get next API call group ID"""
+    with get_db_connection() as con:
+        id = sql_query_single_col(con, "SELECT nextval('api_call_group_id_seq')")[0]
+    return id
+
+
+# maybe might use this later, but for now going to track sessions in memory only
+@router.post("/start-session")
+def start_sessionNOT_USING(body: Dict[str, str] = Body(...)) -> Dict[str, int]:
```
joeflack4, line 380:
```suggestion
def start_session_NOT_USING(body: Dict[str, str] = Body(...)) -> Dict[str, int]:
```

backend/routes/db.py
```diff
@@ -343,16 +275,164 @@ async def get_concepts_post_route(request: Request, id: Union[List[str], None] =
...
+    with get_db_connection() as con:
+        session_id = next_api_call_group_id();
+        q = text("""INSERT INTO sessions (session_id, screen_log) VALUES (:session_id, ARRAY[:page_url])""")
+        run_sql(con, q, {"session_id": session_id, "page_url": page_url})
+    return {"session_id": session_id, "page_num": 1}
+
+
+# maybe might use this later, but for now going to track sessions in memory only
+@router.post("/continue-session")
+def continue_sessionNOT_USING(body: Dict[str, str] = Body(...)):
```
joeflack4, line 391:
```suggestion
def continue_session_NOT_USING(body: Dict[str, str] = Body(...)):
```

backend/utils.py
```diff
@@ -18,10 +20,55 @@
...
+        execution until after `wait` seconds
+        have elapsed since the last time it was invoked.
+        from https://chat.openai.com/share/7ed01b5b-97bd-436a-b23b-542f068a4302"""
+    def decorator(func):
+        def debounced(*args, **kwargs):
+            def call_it():
+                func(*args, **kwargs)
+            if hasattr(debounced, '_timer'):
+                debounced._timer.cancel()
+            debounced._timer = threading.Timer(wait, call_it)
```
joeflack4, line 34: This is interesting. I have no idea how this works. It's weird to see the name of the function referenced inside of itself.
Otherwise looks familiar to how I've done decorators before.

I don't see this being used yet, so I'm guessing it might not be fully functional yet?

backend/utils.py
```diff
@@ -96,6 +143,7 @@ def call_github_action(
     return response
 
 
+# TODO: I don't think we're using this. Delete?
```
joeflack4, line 146: I left it here as a reminder, but yeah that's fine. I have an alternative method for alerting us anyway, which I haven't implemented, but it involves purposefully failing actions.

backend/utils.py
```diff
@@ -158,6 +206,7 @@ def pdump(o):
     print(dump(o))
 
 
+# TODO: I don't think we're using this. Delete?
```
joeflack4, line 209: Sure. I forgot about this.

backend/utils.py
```diff
@@ -175,6 +224,7 @@ def get_nested_from_dict(d: Dict, key_path: List):
     return reduce(operator.getitem, key_path, d)
 
 
+# TODO: I don't think we're using this. Delete?
```
joeflack4, line 227: Sure. I don't remember how this was being used.

backend/utils.py
```diff
@@ -208,17 +258,7 @@ async def decorated_func(*args, **kwargs):
...
-#
-#     inject_to_avoid_circular_imports('get_concepts', get_concepts)
-#     inject_to_avoid_circular_imports('CON', CON)
-#
-# INJECTED_STUFF = {}
-# def inject_to_avoid_circular_imports(name, obj):
-#     INJECTED_STUFF[name] = obj
-
-
+# TODO: never got these working, I think. Would be good to have, though. Not a high priority, but @joeflack4, look at some time
```
joeflack4, line 261: Does this apply to this following issue? If so, I'll add a note about it:
- #637

frontend/src/App.jsx
```diff
@@ -185,20 +185,33 @@ function App(props) {
   useEffect(() => {
     (async () => {
       // start or continue session on server
-      const page_url = urlWithSessionStorage({compressed: true});
+      /*  OLD VERSION, uses start_sessionNOT_USING and continue_sessionNOT_USING:
```
joeflack4, line 188:
```suggestion
      /*  OLD VERSION, uses start_session_NOT_USING and continue_session_NOT_USING:
```

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const dataGetter = useDataGetter();
+  const [concepts, setConcepts] = useState([]);
+  const [loading, setLoading] = useState(false);
+  const totalRows = concept_ids.length;
+  const [perPage, setPerPage] = useState(10);
+
+  const fetchConcepts = async page => {
+    setLoading(true);
+    let ids = concept_ids.slice(page - 1, perPage);
+    let conceptLookup = await dataGetter.fetchAndCacheItems(dataGetter.apiCalls.concepts, ids);
```
joeflack4, line 102: ## 2 trips instead of 1?
I like this pattern of fetching IDs first, then getting full concept objects in theory. It makes more sense on the backend though, as 2 calls is more network traffic. 

1. concept_ids: `        const r = await dataGetter.fetchAndCacheItems(dataGetter.apiCalls.concept_search, searchText)`
2. concepts:     let conceptLookup = await dataGetter.fetchAndCacheItems(dataGetter.apiCalls.concepts, ids);

Did you consider doing this via 1 call instead? Is it inadvisable for any reasons?

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+  );
+}
+
+function ConceptTable(props) {
+  let {concept_ids} = props;
+  const dataGetter = useDataGetter();
+  const [concepts, setConcepts] = useState([]);
+  const [loading, setLoading] = useState(false);
+  const totalRows = concept_ids.length;
+  const [perPage, setPerPage] = useState(10);
```
joeflack4, line 97:
```suggestion
  const [perPage, setPerPage] = useState(30);
```
Maybe set it higher?
Though I guess while we're developing it, doing it in plain orders of magnitude 10, 100, etc, is easier to reason about.

frontend/src/components/AddConcepts.tsx
joeflack4: ## Does a subsequent call of this return from cache?
I get that the first time you call it, it is "fetching and caching". But the second time you call it w/ the same params, does it just return from the cache?

frontend/src/components/AddConcepts.tsx
joeflack4: I guess it looks like there's a way for the user to change how many rows there are as well?

frontend/src/state/AppState.jsx
```diff
@@ -251,19 +251,32 @@ export function unabbreviateDefinitions(defs) {
   }
   return definitions;
 }
-export function urlWithSessionStorage(newCset) {
+export function getSessionStorage() {
```
joeflack4, line 254: ## Frontend session storage?
I think you mentioned something about this the other day. It looks like this isn't currently being utilized? Is this planned?
If we use it, how does it benefit us?
How is this session storage different than how we're already doing frontend caching?

backend/db/ddl-21-sessions.jinja.sql
```diff
@@ -0,0 +1,23 @@
...
```
joeflack4, line 1: I can't remember, are we currently using this table?

I think you said maybe not? I saw you were doing "`next_api_call_group_id()`" to get session IDs.

backend/db/utils.py
joeflack4: Is this still being used, or part of an abandoned alternative method?

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
+import React, {useState, useEffect} from 'react';
```
joeflack4, line 1: ## Remove `Graph` route for now?
Screenshot 2024-04-26 at 6 32 43 PM

joeflack4, line 1: ## Occasionally getting harmless console errors - couldn't replicate
When I first loaded the app in local, I got a bunch of console errors, but I didn't screenshot.

Then, I refreshed the frontend and went back to the add-concepts page, and got `DataGetter.jsx:107 Uncaught (in promise) Error: AxiosError: Request failed with status code 422`

<details><summary>Screenshot</summary>
<p>

Screenshot 2024-04-26 at 6 34 01 PM

</p>
</details> 

In neither case was did my experience as a user result in any noticeable issue, though.

After I restarted the backend and frontend again, I got no more console errors, interestingly enough.

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+      // setting min lenght for searchText
+      if (searchText.length >= 3) {
+        // updating the ref variable with the current searchText
+        lastRequest.current = searchText;
+        const r = await dataGetter.fetchAndCacheItems(dataGetter.apiCalls.concept_search, searchText)
+        if (lastRequest.current === searchText) {
+          console.log("response is valid!", r);
+          setConceptIds(r);
+        } else {
+          console.log("discarding api response", searchText, lastRequest.current);
```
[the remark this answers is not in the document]
joeflack4: Actually never mind. I see that this is going to happen very often / all the time.

For example, I typed 'infarction' and got this:
```
discarding api response infarcti infarction
AddConcepts.tsx:72 discarding api response infarct infarction
AddConcepts.tsx:72 discarding api response infarc infarction
AddConcepts.tsx:72 discarding api response inf infarction
AddConcepts.tsx:72 discarding api response infar infarction
AddConcepts.tsx:72 discarding api response infa infarction
AddConcepts.tsx:72 discarding api response infarctio infarction
```

Perhaps when we no longer need this for troubleshooting, we should not console log this at all, I think.

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const _concepts = ids.map(id => conceptLookup[id]);
+    setConcepts(_concepts);
+    setPerPage(newPerPage);
+    setLoading(false);
+  };
+  useEffect(() => {
+    fetchConcepts(1); // fetch page 1 of users
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [concept_ids]);
+  console.log(concepts);
```
joeflack4, line 123: Do we still need this, or was this for troubleshooting?

FYI, it's printing out a lot. It printed 10+ times just when I did my search for 'infarction', before I went through any pages.

<p>

```
api_call_group_id 20417
MuiAppBar.jsx:95 null
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:72 discarding api response infarcti infarction
AddConcepts.tsx:72 discarding api response infarct infarction
AddConcepts.tsx:72 discarding api response infarc infarction
AddConcepts.tsx:72 discarding api response inf infarction
AddConcepts.tsx:72 discarding api response infar infarction
AddConcepts.tsx:72 discarding api response infa infarction
AddConcepts.tsx:72 discarding api response infarctio infarction
AddConcepts.tsx:69 response is valid! Array(1518)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(0)
AddConcepts.tsx:123 Array(10)
```

</p>
</details>

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const _concepts = ids.map(id => conceptLookup[id]);
+    setConcepts(_concepts);
+    setLoading(false);
+  };
+  const handlePageChange = page => {
+    fetchConcepts(page);
+  };
+  const handlePerRowsChange = async (newPerPage, page) => {
+    setLoading(true);
+    let ids = concept_ids.slice(page - 1, perPage);
```
joeflack4, line 112: ## Bug fix: incorrect, vanishing results on page change
```suggestion
    let ids = concept_ids.slice(page - 1, page - 1 + perPage);
```

Without the fix, I was seeing a situation where going to the next page didn't show any new concepts, but simply shortened the list of concepts already showing.

Page 1: 10 concepts


Page 10: 1 concept (from the same original list)

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+  },
+  {
+    selector: row => row.domain_id,
+    name: "Domain",
+    sortable: true,
+    width: 200,
+  },
+  {
+    selector: row => row.total_cnt,
+    name: "Total Count",
```
joeflack4, line 36:
```suggestion
    name: "Record Count",
```

joeflack4: Also maybe for these column headers, we want to capitalize only the first word?

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+    setPerPage(newPerPage);
+    setLoading(false);
+  };
+  useEffect(() => {
+    fetchConcepts(1); // fetch page 1 of users
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [concept_ids]);
+  console.log(concepts);
+  return <DataTable
+            title="Users"
```
joeflack4, line 125:
```suggestion
            title="Concepts"
```

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+  readonly invalid_reason: string;
+  readonly domain_cnt: number;
+  readonly domain: string;
+  readonly total_cnt: number;
+  readonly distinct_person_cnt: string;
+}
+
+const columns = [
+  {
+    name: "Concept Name",
```
joeflack4, line 23: ## Display full concept name
Right now if the concept name is too large, it gets truncated (...).

Screenshot 2024-04-26 at 6 46 51 PM

Perhaps when we add the "add concept" UI, then when the user selects the concept row, we can have it show the full name and that will be sufficient?

In addition, we could add an `onHover` tooltip as well, but maybe the former is sufficient.

frontend/src/components/AddConcepts.tsx
joeflack4, line 1: ## Everything in issue complete?
Is this PR no longer a draft; ready to merge?

Let's make sure everything in this issue is either (a) addressed by this PR, (b) will be addressed later, or (c) no longer relevant:
- #601

frontend/src/components/AddConcepts.tsx
Sigfried: I should just wrap text like happens in the comparison table

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,134 @@
...
+          setConceptIds(r);
+        } else {
+          console.log("discarding api response", searchText, lastRequest.current);
+        }
+      }
+    })();
+  }, [searchText]);
+
+  return (
+    <div className="App" style={{paddingLeft: 100}}>
```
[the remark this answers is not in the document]
joeflack4: We can actually get rid of the `className` in these cases.

backend/db/utils.py
```diff
@@ -206,9 +226,30 @@ def refresh_derived_tables(
...
+# @joeflack4: finally put in connection pooling. (it didn't help the problem I was having)
+# got the code from our friend: https://chat.openai.com/share/0472ba44-0e58-44ad-bbcb-5eabffae3d39
+def set_search_path(dbapi_connection, connection_record):
+    """Set the search_path when connecting."""
+    if SCHEMA:
+        cursor = dbapi_connection.cursor()
+        cursor.execute(f"SET SESSION search_path='{SCHEMA}'")
+        cursor.close()
+
+def get_db_connection(isolation_level='AUTOCOMMIT', schema: str = SCHEMA, local=False) -> Connection:
```
joeflack4, line 238: ## Connection pooling
@Sigfried Was this necessary at this point or a nice-to-have?

backend/db/utils.py
```diff
@@ -50,6 +52,24 @@
 SCHEMA = CONFIG["schema"]
 
 
+def log_concepts(session_id: int, concept_ids: List[int]):
```
joeflack4, line 55: ## Log concepts
Just curious - it says 'log', but it doesn't look like it's logging; it's adding to a session. Should this be renamed, like `add_concepts_to_session()`?

frontend/src/state/AppState.jsx
```diff
@@ -77,17 +95,48 @@ export function useAlertsDispatch() {
...
+    // let {collapsePaths, // collapsedDescendantPaths,
+    //   nested, hideRxNormExtension, hideZeroCounts} = {...unpersistedDefaultState, ...state};
+    const { graphOptions, gc } = action;
+    // real kludgy passsing gc in, but trying not to rewrite a bunch more code than I need to
     switch (action.type) {
+      case "graphOptions": {
+        return {...state, graphOptions};
+      }
+      // took this from GraphState, but it still stores graphOptions state there as well as here
+      case 'TOGGLE_NODE_EXPANDED':
```
joeflack4, line 120: Is this the code here and below to address?:
- #787

frontend/src/state/StorageProvider.jsx
```diff
@@ -0,0 +1,222 @@
+import React, {createContext, useContext, useState, useCallback, useEffect} from "react";
+import {createSearchParams, useSearchParams, /* useLocation, Navigate, */ } from "react-router-dom";
+import {isEmpty, omit} from "lodash";
+
+// starting from https://chat.openai.com/share/6cc19197-8a46-49ce-9100-84a83d0d2bf1
+// at least for now, moving all search params to sessionStorage
+export function useSessionStorage() {
```
joeflack4, line 7: Siggie wrote:
> Made it so any state that is added to the querystring will be added to sessionStorage

This sounds good to me.

frontend/src/components/AddConcepts.tsx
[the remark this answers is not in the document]
joeflack4: Siggie may have left it as `App` on purpose. Might be some styles

frontend/src/state/GraphState.jsx
```diff
@@ -490,70 +508,84 @@ export class GraphContainer {
...
-        this.options.specialConceptTreatment[type] = row.specialTreatment;
-      }
+        // this.options.specialConceptTreatment[type] = row.specialTreatment;
+        // graphOptions.specialConceptTreatment[type] = row.specialTreatmentDefault;
+      // }
       rows[type] = row;
     }
-    this.statsOptions = rows;
+    this.graphDisplayConfig = rows;
+    this.graphDisplayConfigList = sortBy(rows, d => d.displayOrder);
```
joeflack4, line 531: ## Row display order
@Sigfried Saw in [this commit](https://github.com/jhu-bids/TermHub/commit/050067ce41f9bb06fae763c03a3f3882669058a8). Is this change about sorting by column?

frontend/src/components/AddConcepts.tsx
```diff
@@ -0,0 +1,395 @@
...
+  return (
+    <div style={{paddingLeft, paddingRight, width: divWidth}}>
+      <h1>Concept Search</h1>
+      <input style={{width: 350}} type="text" placeholder="match characters in concept name"
+             onChange={(e) => setSearchText(e.target.value)}
+             value={searchText}
+             autoFocus={true}
+      />
+      {'\u00A0'}{'\u00A0'}{'\u00A0'}{found_concept_ids.length ? found_concept_ids.length.toLocaleString() + ` concepts match "${searchText}"` : ""}
+      ;{'\u00A0'}{hiddenMatches.length ? hiddenMatches.length.toLocaleString() + ' already included and not listed' : ""}
```
joeflack4, line 153:
```suggestion
      {'\u00A0'}{hiddenMatches.length ? hiddenMatches.length.toLocaleString() + ' already included and not listed' : ""}
```
I think this semicolon here is manifesting in:
Screenshot 2024-05-29 at 4 38 06 PM

frontend/src/state/AppState.jsx
```diff
@@ -55,39 +48,70 @@ export function useCodesetIds() {
...
+  let state = storageProvider.getItem('cids') || [];
 
-const AlertsContext = createContext(null);
-const AlertsDispatchContext = createContext(null);
-export function AlertsProvider({ children }) {
-  const [alerts, dispatch] = useReducer(alertsReducer, {});
-
+  const dispatch = action => {
+    let latestState = storageProvider.getItem('cids') || [];
+    const stateAfterDispatch = cidsReducer(latestState, action);
```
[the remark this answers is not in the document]
joeflack4: Siggie: The query string is in flux. In the end, these added concepts won't be going on the query string anyway.

frontend/src/state/AppState.jsx
```diff
@@ -114,11 +138,108 @@ export function useHierarchySettings() {
...
+  name: 'graphOptions', reducer: graphOptionsReducer, initialSettings: {}, storageProviderGetter: useSearchParamsState, });
+
+function makeProvider({name, reducer, initialSettings, storageProviderGetter, jsonify=false, forceUpdateForConsumers=false}) {
+  const nonUpdatingReducer = (state, action) => {
+    const newState = reducer(state, action);
+    return isEqual(state, newState) ? state : newState;
+  };
+
+  const Context = createContext();
+  const Provider = ({children}) => {
```
joeflack4, line 190: Are `Context` and `Provider` new? 
Are you ready to discuss these or are they still in flux?

This is the first time I've seen `Context` or `Provider` by themselves, without some qualifier before it, like `MySuchAndSuchContext`.
I don't see how they're used yet either. Are they being used yet?

Sigfried:
Pooling: I thought it would help with a slowness problem, but it didn't. We
decided to keep it because it could prevent memory overloads at some point.
Log concepts: I don't know if that's being used now

backend/db/utils.py
```diff
@@ -218,9 +220,31 @@ def refresh_derived_tables(
...
+def set_search_path(dbapi_connection, connection_record):
+    """Set the search_path when connecting."""
+    if SCHEMA:
+        cursor = dbapi_connection.cursor()
+        cursor.execute(f"SET SESSION search_path='{SCHEMA}'")
+        cursor.close()
+
+def get_db_connection(isolation_level='AUTOCOMMIT', schema: str = SCHEMA, local=False) -> Connection:
+    """Get database connection object"""
+    if schema == None:
```
joeflack4, line 234: ## Bug fixes: To new implementation of `get_db_connection()`

### Connecting to public schema
Siggie recently added a `ValueError` for `if not schema` in `get_db_connection()`. However, we have like 15 instances where we are connecting via `schema=''`. This is how we connect to `public`, I thought. I checked w/ ChatGPT and they say that `''` and `'public'` are not the same. However I'm not sure if I trust their response. If it ain't broke, don't fix it; so I want to keep allowing this.

@Sigfried Is there something that broke because schema was explicitly set to `None` or `''`? If so, then I have to maybe change all these instances to `'public'` and run the test suite again (which should hopefully show that nothing has broke).

## Schema not getting passed around correctly
`set_search_path()` was using `SCHEMA`, and didn't have a `schema` arg either.

## Refactor back to old method if we need
@Sigfried said that this new method didn't help them, so we can refactor back if needed.

test/test_database.py
```diff
@@ -33,7 +33,7 @@ def setUpClass(cls):
     def test_all_enclave_csets_in_termhub(self):
```
joeflack4, line 33: ## Bug fix: `test_all_enclave_csets_in_termhub()`

test/test_backend/db/test_utils.py
```diff
@@ -25,17 +25,27 @@ class FetchAuditTestRunner(unittest.TestCase):
...
+        Normally this would only need to be done at the end in tearDown(), but in cases of errors, rows could sneak in
+        and not be removed at the end, so we should also clean in setUp()."""
+        run_sql(con, f"DELETE FROM fetch_audit WHERE comment LIKE 'Unit testing.%';")
+
+
     @classmethod
     def setUpClass(cls):
         """setUp"""
         cls.con = get_db_connection(schema='')
+        cls.clean_fetch_audit(cls.con)
```
joeflack4, line 41: ## Bug fix: Edge cases where errors cause permanent state change, causing tests to fail
During some tests, temporary rows get inserted into `fetch_audit`. They're removed during `tearDown()`. However, in cases where there were errors during a run, `tearDown()` doesn't get to run, leaving an unclean state in which the tests would continue to fail indefinitely. I've added this cleanup step to the `setUp()` now.

frontend/package.json
```diff
@@ -35,14 +35,16 @@
...
     "papaparse": "^5.4.1",
-    "react": "^18.2.0",
+    "react": "^18.3.1",
     "react-data-table-component": "^7.5.4",
-    "react-dom": "^18.2.0",
+    "react-dom": "^18.3.1",
     "react-draggable": "^4.4.6",
     "react-hooks-global-state": "^2.1.0",
     "react-inspector": "^6.0.2",
+    "react-markdown": "^9.0.1",
```
joeflack4, line 44: Markdown for state overview page, etc? Cool

frontend/src/components/AddConcepts.tsx
joeflack4, line 1: ## Jest tests
- [ ] @Sigfried Include at least some tests (Jest at least)

joeflack4, line 1: ## Wrap text instead of `...`
- [ ] @Sigfried  Concept label: Ensure the full label shows and is not truncated with .... Can just wrap text like happens in the comparison table.

Sigfried: I need to switch to working on comparison rpt stuff for the next vocab update. Will have to work on the tests after that I think, or at least after merging this pull request.

PR #755, opened 2024-04-13
Refresh button fixes

joeflack4:
## Changes
Refresh button fixes
- Bug fix: Refresh button. I think was broken cuz was passing params that didn't exist on backend
- Update: Now runs GH action for refresh instead of running on server.

### DIFF
--- a/backend/routes/db.py
+++ b/backend/routes/db.py
@@ -19,7 +19,7 @@ from backend.api_logger import Api_logger, get_ip_from_request
 from backend.db.refresh import refresh_db
 from backend.db.queries import get_concepts
 from backend.db.utils import get_db_connection, sql_query, SCHEMA, sql_query_single_col, sql_in, sql_in_safe, run_sql
-from backend.utils import return_err_with_trace, commify
+from backend.utils import call_github_action, return_err_with_trace, commify
 from enclave_wrangler.config import RESEARCHER_COLS
 from enclave_wrangler.models import convert_rows
 from enclave_wrangler.objects_api import get_n3c_recommended_csets, get_concept_set_version_expression_items, \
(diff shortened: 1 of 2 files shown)


backend/routes/db.py
```diff
@@ -365,21 +365,9 @@ def get_researchers(ids: List[str] = Query(...), fields: Union[List[str], None]
...
-     If we want to make changes to the refresh, we have to redeploy the whole app to make that work, rather than pushing to develop.
-     4. Uses more server resources.
-     5. Possible server stability issues
-     I'm not sure, but I wonder if there is some edge case where an error that happens during the refresh, or other unanticipated side effects, could have an effect on performance or stability of the web server."""
-    # response: Response = call_github_action('refresh-db')
-    # return response
-    refresh_db()
+    """Triggers refresh of the database"""
+    response: Response = call_github_action('refresh-db')
+    return response.status_code
```
joeflack4, line 370: ## Return: status code
Nothing actually is done with the return value on the frontend, but I think it's better to return this or something else rather than the whole response obj.

backend/routes/db.py
```diff
@@ -365,21 +365,9 @@ def get_researchers(ids: List[str] = Query(...), fields: Union[List[str], None]
 
 @router.get("/db-refresh")
 def db_refresh_route():
-    """Triggers refresh of the database
-    todo: May want to change this back to GH action for reasons:
```
joeflack4, line 369: ## Rationale
Wrote a bunch of comments here for these reasons.
I think another one is 'server memory', but I don't think that was listed there.

backend/routes/db.py
```diff
@@ -365,21 +365,9 @@ def get_researchers(ids: List[str] = Query(...), fields: Union[List[str], None]
...
-     3. Harder to make changes
-     If we want to make changes to the refresh, we have to redeploy the whole app to make that work, rather than pushing to develop.
-     4. Uses more server resources.
-     5. Possible server stability issues
-     I'm not sure, but I wonder if there is some edge case where an error that happens during the refresh, or other unanticipated side effects, could have an effect on performance or stability of the web server."""
-    # response: Response = call_github_action('refresh-db')
-    # return response
-    refresh_db()
+    """Triggers refresh of the database"""
+    response: Response = call_github_action('refresh-db')
```
joeflack4, line 369: ## The crux
Calls action instead of run on backend.

frontend/src/components/AboutPage.jsx
```diff
@@ -109,7 +109,7 @@ export function AboutPage() {
       console.log('Triggering database refresh and clearing cache so new data will be fetched when ready');
       // empty cache immediately, and then again after the db-refresh call is done
       dataCache.emptyCache();
-      await dataGetter.axiosCall('db-refresh', {backend: true, verbose: false, title: 'Refreshing VS-Hub database from Enclave'});
```
joeflack4, line 112: ## Removed superfluous args
I'm not sure why these args were here. The refresh route doesn't have any params.

PR #756, opened 2024-04-14
Bug fixes: Discarded drafts & `fetch_failures_0_members`

joeflack4:
## Changes
    - Bug fix: Discarded drafts cause error in fetch failures 0 members
    - Bug fix: Runs fetch failures (if needed) in every instance of normal DB refresh running, even if the DB refresh detects that it does not need to run.

### DIFF
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -23,6 +23,17 @@ DESC = 'Refresh TermHub database w/ newest updates from the Enclave using the ob
 SINCE_ERR = '--since is more recent than the database\'s record of last refresh, which will result in missing data.'
 
 
+def trigger_resolve_failures(
+    resolve_fetch_failures_excess_items=False, resolve_fetch_failures_0_members=True, local=False
+):
+    """Trigger the GitHub actions to resolve fetch failures"""
+    # Routine: Check for and resolve any open fetch failures
+    if resolve_fetch_failures_excess_items:
+        call_github_action('resolve-fetch-failures-excess-items')
+    if resolve_fetch_failures_0_members:
+        resolve_failures_0_members_if_exist(local)
+
+
 # todo: low priority: track the time it takes for this process to run, and then update the `manage` table, 2 variables:
 #  total time for downloads, and total time for uploading to db (perhaps for each table as well)
 # todo: store runs of db refresh in a table, so can have a reference for troubleshooting. e.g. date of run, and what
@@ -57,6 +68,7 @@ def refresh_db(
     if is_refresh_active():
         print('INFO: Refresh already in progress. When that process completes, it will restart again. Exiting.')
         update_db_status_var('new_request_while_refreshing', start_time, local)
+        trigger_resolve_failures(resolve_fetch_failures_excess_items, resolve_fetch_failures_0_members, local)
         return
     # end_time: Even though in reality the refresh will not end 1 microsecond after the start time, we're setting it
     # this way because this is the easiest way to make sure that future refreshes will not miss any new data that was
@@ -94,11 +106,7 @@ def refresh_db(
             # Update status vars
             update_db_status_var('last_refresh_exited', current_datetime(), local)
             update_db_status_var('refresh_status', 'inactive', local)
-            # Routine: Check for and resolve any open fetch failures
-            if resolve_fetch_failures_excess_items:
-                call_github_action('resolve-fetch-failures-excess-items')
-            if resolve_fetch_failures_0_members:
-                resolve_failures_0_members_if_exist(local)
+            trigger_resolve_failures(resolve_fetch_failures_excess_items, resolve_fetch_failures_0_members, local)
 
     if new_data:
         print('DB refresh complete.')
--- a/backend/db/resolve_fetch_failures_0_members.py
+++ b/backend/db/resolve_fetch_failures_0_members.py
@@ -6,6 +6,7 @@ import os
 import sys
 import time
 from argparse import ArgumentParser
+from copy import copy, deepcopy
 from datetime import datetime
 from typing import Dict, List, Set, Tuple
 
@@ -74,7 +75,8 @@ def resolve_fetch_failures_0_members(
      small performance gain."""
     print("Resolving fetch failures: ostensibly new (possibly draft) concept sets with >0 expressions but 0 members")
     # Collect failures
-    failed_cset_ids, failure_lookup = get_failures_0_members(version_id, use_local_db)
+    failed_cset_ids_prefilter, failure_lookup = get_failures_0_members(version_id, use_local_db)
+    failed_cset_ids = copy(failed_cset_ids_prefilter)
     if not failed_cset_ids:
         print("No failures to resolve.")
         return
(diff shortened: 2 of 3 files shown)


backend/db/refresh.py
```diff
@@ -57,6 +68,7 @@ def refresh_db(
     if is_refresh_active():
         print('INFO: Refresh already in progress. When that process completes, it will restart again. Exiting.')
         update_db_status_var('new_request_while_refreshing', start_time, local)
+        trigger_resolve_failures(resolve_fetch_failures_excess_items, resolve_fetch_failures_0_members, local)
```
joeflack4, line 71: ## Always run fetch failures if needed
Bug fix: Now, if the refresh doesn't need to run, it still triggers resolution of fetch failures.

This fix is good because when 0 members occur because of a new draft, the members might not be ready by the end of the refresh, and it'd be good if the next time the refresh ran, it tried to pick them up.

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -93,6 +93,9 @@ def resolve_fetch_failures_0_members(
         # Fetch data
         csets_and_members: Dict[str, List[Dict]] = fetch_cset_and_member_objects(
             codeset_ids=list(failed_cset_ids), flag_issues=False)
+        # TODO: mark discarded drafts resolved
```
joeflack4, line 96: ## TODO: mark discarded drafts resolved
Otherwise, `fetch_failures_0_members` will keep running w/ every refresh (every 20 min) because it will think that the discarded draft is still an issue.

I'm going to do this in a new PR.

enclave_wrangler/objects_api.py
```diff
@@ -464,7 +464,19 @@ def fetch_cset_and_member_objects(
...
+
+    # - handle discarded drafts
+    if codeset_ids and len(cset_versions) < len(codeset_ids):
+        discarded_draft_ids: Set[int] = \
+            set(codeset_ids).difference([x['properties']['codesetId'] for x in cset_versions])
+        # todo: could confirm is marked a draft in TermHub db for extra confidence, and update this message
+        logging.warning(
+            'Tried to fetch data for the following cset versions, but the Enclave returned with no data. It is '
+            'probably the case that these are discarded drafts. Offending cset ids: ' +
+            ', '.join([str(x) for x in discarded_draft_ids]))
```
joeflack4, line 477: ## Logs about discarded draft

enclave_wrangler/objects_api.py
```diff
@@ -464,7 +464,19 @@ def fetch_cset_and_member_objects(
         cset_versions: List[Dict] = fetch_objects_since_datetime('OMOPConceptSet', since, verbose) or []
     # - Filter any old drafts from sourceApplicationVersion 1.0. Old data model. No containers, etc. See also:
     #  https://github.com/jhu-bids/TermHub/actions/runs/6489411749/job/17623626419
-    cset_versions = [x for x in cset_versions if x['properties']['conceptSetNameOMOP']]
+    cset_versions = [x for x in cset_versions if x and x['properties']['conceptSetNameOMOP']]
```
joeflack4, line 467: ## Bug fix: Filtering out discarded drafts
Otherwise, would cause a `KeyError`.

joeflack4, line 467: ## todo?: Best way to return info about discarded drafts?
The way that this function `fetch_cset_and_member_objects()` returns information about discarded drafts is purely implicit. If it receives `codeset_ids` to fetch, then if the draft is discarded, that cset will simply not be returned.

The fetch failures script knows about this and handles accordingly. But perhaps doing this in some more explicit way, such as adding a `discarded_drafts` key to the object/dict returned by the func would be nice.

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -93,23 +95,36 @@ def resolve_fetch_failures_0_members(
...
+
+        # - identify discarded drafts
+        # TODO: ascertain discarded drafts
+        #  - how to ascertain?
+        #    - 1 (ok~): detect discarded drafts as not returned from fetch_cset_and_member_objects(). ~confirm w/ amin et al
+        #    - 2 (x): Ask Amin et al: is the enclave API reliable for discarded drafts? Some of these are appearing in the
+        #    enclave API but not in the UI. x: I think (2) is not a case I need to worry about. i think they weren't discared.
+        #    - This line below should be enough to handle. Just want to verify (1)
+        discarded_cset_ids: Set[int] = set([x for x in failed_cset_ids if x not in csets_and_members['OMOPConceptSet']])
+        # TODO: mark discarded drafts resolved
```
joeflack4, line 107: ## FYI: Investigation
FYI, in this run that I looked at (new data from doing a refresh since Jan 1st), there is:

20 fetch failures
- 1 empty return (probably discarded draft)
- 19 returned from `fetch_csets_and_members()`
  - 2 success cases (members fetched)
  - 17 persistent drafts (appear in TermHub & in enclave API but not in enclave UI)

joeflack4, line 107: ## @Sigfried Why empty csets have related csets?
Example:
why empty cset has related

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -93,23 +95,36 @@ def resolve_fetch_failures_0_members(
...
+        # _report_success(discarded_cset_ids, failure_lookup, 'Success result: Discarded draft', use_local_db)
+
+        # - identify persistent drafts
```
joeflack4, line 110: ## How to handle persistent drafts?
I think I'll make an issue for this.

I think there are drafts that will remain drafts permanently. With the current setup, we're just going to keep trying to fetch their members every 20 minutes forever.

Perhaps I can update `fetch_audit` with a new field `last_fetch_attempt`, and for drafts that are old (1 day; 1 week?) we check less often? How often? 1x/day? It will be annoying though for someone who marks their draft resolved after a few weeks and wants to see it in TermHub immediately. Perhaps the way around this is if there's some way for the refresh GH action to know if it was triggered manually, rather than on cronjob, then it should automatically trigger the `fetch_failures_0_members` for these persistent drafts as well.

backend/db/resolve_fetch_failures_0_members.py
joeflack4, line 107: ## Question: Run `refresh --since` longer than Jan 1?
I got 20 new fetch failures because, after fixing a bug in which drafts were being erroneously marked as resolved (we would not fetch their members), I ran this since the last time I remember running a big refresh (some time in January). However, the last time we initialized was last June or July I think. Maybe I need to do `refresh --since` back to that time, just in case? I suppose so.

joeflack4: - #758

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -74,7 +75,8 @@ def resolve_fetch_failures_0_members(
      small performance gain."""
     print("Resolving fetch failures: ostensibly new (possibly draft) concept sets with >0 expressions but 0 members")
     # Collect failures
```
joeflack4, line 77: ## Data fix: "Success result: No members after 2 hours. Considering resolved."
We had previously set up this script to run for 2 hours and, if no members were detected, we were considering it to be the case that there was simply no expansion; that there were 0 members. However, I believe that this erroneously did not check to see if those csets were drafts, and thus erroneously marked them as being completed.

I solved by running a query to remove the "success" part of the comment string and set `success_datetime` to NULL. Then I fetched data for all of these csets.

backend/db/resolve_fetch_failures_0_members.py
joeflack4: I created an issue:
- #759

backend/db/resolve_fetch_failures_0_members.py
joeflack4: I decided to do this. The last time we initialized was `2023-07-12T03:51:48.292700-04:00`, so I decided to run this since `2023-07-11`. I also decided it was time to be able to run this via GitHub actions, so I created:
- #760

joeflack4: I ran it, but I was surprised to see [very few deltas](https://github.com/jhu-bids/TermHub/blob/develop/docs/backend/db/analysis.md).

It looks like it was able to fetch many long-lived drafts (which we had previously erroneously marked finalized): 37 new ones in this run.

PR #763, opened 2024-04-21
Bugfix: GH action triggering

joeflack4:
## Changes
Bugfix: GH action triggering
- Bugfix: PAT: This won't be represented in the code, but the personal access token for doing this was not included in the .env.
- Update: Now call_github_action() raises RuntimeError if error received from POST.

### DIFF
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -92,7 +92,13 @@ def call_github_action(
         if params:
             payload["inputs"] = params
 
-    response = post(url, headers=headers, data=json.dumps(payload))
+    response: Response = post(url, headers=headers, data=json.dumps(payload))
+    if response.status_code >= 400:
+        raise RuntimeError(
+            f"Error calling GitHub action {action_name} with params:"
+            f"\n{params}."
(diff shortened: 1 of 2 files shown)


backend/utils.py
```diff
@@ -92,7 +92,13 @@ def call_github_action(
         if params:
             payload["inputs"] = params
 
-    response = post(url, headers=headers, data=json.dumps(payload))
+    response: Response = post(url, headers=headers, data=json.dumps(payload))
+    if response.status_code >= 400:
+        raise RuntimeError(
+            f"Error calling GitHub action {action_name} with params:"
```
joeflack4, line 98: ## Now raises error
If error came back from GitHub API. I thought I was already doing this; guess not.

joeflack4, line 98: ## The fix
The `GH_LIMITED_PERSONAL_ACCESS_TOKEN` variable was missing from `.env`. This is the result of Joe's computer dying in October, and Joe having used Siggie's `.env` since then. At some point, the `.env` of TermHub was updated to match this, and at that point is when these GH action calls stopped working.

PR #764, opened 2024-04-21
Remove refresh hours buffer

joeflack4:
## Changes
Update: DB Refresh: Remove 2 hour refresh buffer. Now no buffer. Just runs since the last time it exited.
- Update: Also made clarifying comment and cleaned up code to make 'end_time' more clear. Also changed its name to 'end_time_reported'.

### DIFF
--- a/backend/db/refresh.py
+++ b/backend/db/refresh.py
@@ -27,7 +27,7 @@ HELP_SINCE = """A timestamp by which new data should be fetched. If not present,
 refreshed and fetch new data from that time. Valid formats: 
 - Simple date timestamp, YYYY-MM-DD, e.g. 2022-02-22.
 - ISO 8601 datetime with timezone offset, YYYY-MM-DDTHH:MM:SS.SSSSSS+HH:MM, e.g. 2022-02-22T22:22:22.222222+00:00."""
-DEFAULT_BUFFER_HOURS = 2
+DEFAULT_BUFFER_HOURS = 0
 
 
 def trigger_resolve_failures(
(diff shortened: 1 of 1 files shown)


backend/db/refresh.py
```diff
@@ -27,7 +27,7 @@
 refreshed and fetch new data from that time. Valid formats: 
 - Simple date timestamp, YYYY-MM-DD, e.g. 2022-02-22.
 - ISO 8601 datetime with timezone offset, YYYY-MM-DDTHH:MM:SS.SSSSSS+HH:MM, e.g. 2022-02-22T22:22:22.222222+00:00."""
-DEFAULT_BUFFER_HOURS = 2
+DEFAULT_BUFFER_HOURS = 0
```
joeflack4, line 30: ## The change
**Rationale**: We're running `fetch_failures_0_members` effectively every 20 minutes now, so this buffer is now redundant. The buffer was basically functioning to do the same thing; to find members that were not yet available when the cset was initially fetched.

backend/db/refresh.py
```diff
@@ -97,18 +97,20 @@ def refresh_db(
     """
     local = use_local_db
     print('INFO: Starting database refresh.', flush=True)  # flush: for gh action
+
+    # Calculate: Start & end time
+    # end_time: Even though in reality the refresh will not end 1 microsecond after the start time, we're setting it
+    # this way because this is the easiest way to make sure that future refreshes will not miss any new data that was
+    # added between when the refresh started and ended.
     t0, start_time = datetime.now(), current_datetime()
+    end_time_reported: str = tz_datetime_str(dp.parse(start_time) + timedelta(microseconds=1))
```
joeflack4, line 106: ## Clarification: `end_time`
The `end_time` reported for the refresh is actually the same as the time when it started. This is to ensure that no data is missed the next time the refresh runs.

I updated the variable name, the position where it is declared, and the code comments in order to make this less confusing.

PR #765, opened 2024-04-23
`resolve_fetch_failures_0_members()`: Error only if cset > 2 hours

joeflack4:
## Changes
`resolve_fetch_failures_0_members()` updates
- Bugfix: Error message for failures showing incorrect list
- Update: rename param: runtime --> time_to_expand_threshold
- Update: Now errors out if fails to fetch members for non-draft, but only if it has been finalized longer than time_to_expand_threshold.

### DIFF
--- a/backend/db/resolve_fetch_failures_0_members.py
+++ b/backend/db/resolve_fetch_failures_0_members.py
@@ -19,7 +19,7 @@ from backend.db.resolve_fetch_failures_excess_items import resolve_fetch_failure
 from backend.db.utils import SCHEMA, fetch_status_set_success, get_db_connection, reset_temp_refresh_tables, \
     select_failed_fetches, refresh_derived_tables
 from enclave_wrangler.objects_api import concept_set_members__from_csets_and_members_to_db, \
-    fetch_cset_and_member_objects
+    fetch_cset_and_member_objects, get_csets_over_threshold
 
 DESC = "Resolve any failures resulting from fetching data from the Enclave's objects API."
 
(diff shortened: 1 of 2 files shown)


backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -162,22 +162,29 @@ def resolve_fetch_failures_0_members(
...
               f" members expanded yet: {', '.join([str(x) for x in still_draft_cset_ids])}")
-    true_failed_cset_ids: Set[int] = set(failed_cset_ids) - still_draft_cset_ids
+    non_draft_failure_ids: Set[int] = set(failed_cset_ids) - still_draft_cset_ids
+
+    # Filter by only if has been finalized longer than we would expect it should take for expansion to be available
+    # noinspection PyUnboundLocalVariable
+    still_draft_csets: List[Dict] = [
+        x['properties'] for x in csets_and_members['OMOPConceptSet']
+        if x['properties']['codesetId'] in non_draft_failure_ids]
+    final_failure_ids: Set[int] = get_csets_over_threshold(still_draft_csets, int(expansion_threshold_seconds / 60))
```
joeflack4, line 172: Now errors only if the cset is too old. We figure that 2 hours is a sufficient amount of time to give to the Enclave to expand.

@Sigfried Should we take this time not simply to throw an error, but should we mark as finalized if this happens?
I have a better proposal, though:
`TODO: if every expression is set to `isExcluded=true`, we expect no members. Can add as a check rather than wait 2hrs`

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -162,22 +162,29 @@ def resolve_fetch_failures_0_members(
...
+              f" not actually have members. Reporting resolved: {', '.join([str(x) for x in final_failure_ids])}")
         comment = 'Success result: No members after 2 hours. Considering resolved.'
-        _report_success(failed_cset_ids, failure_lookup, comment, use_local_db)
-    elif true_failed_cset_ids:
+        _report_success(list(final_failure_ids), failure_lookup, comment, use_local_db)
+    elif final_failure_ids:
         raise RuntimeError('Attempted to resolve fetch failures for the following concept sets, but was not able to do '
                            'so. It may be that the Enclave simply has not expanded their members yet:\n\n'
-                           f'{", ".join([str(x) for x in failed_cset_ids])}')
+                           f'{", ".join([str(x) for x in final_failure_ids])}')
```
joeflack4, line 187: Bugfix: was printing the wrong IDs when failed. Was including csets that were long-lived drafts.

enclave_wrangler/objects_api.py
```diff
@@ -365,6 +365,24 @@ def get_bidirectional_csets_sets(con: Connection = None) -> Tuple[Set[int], Set[
     return db_codeset_ids, enclave_codeset_ids
 
 
+def get_csets_over_threshold(
```
joeflack4, line 368: ## Refactor: create `get_csets_over_threshold()`
Now, there are multiple places in the code that are trying to ascertain if a cset is older than a certain period of time. The difficult part of this is making sure that the timestamp is in the same timezone for both datetimes being compared, so it's important to use a common func for this.

PR #786, opened 2024-05-08
Update DB config

joeflack4:
## Changes
- Update DB config w/ missing table relationships
- Update test_dependency_map(): Better error messaging

### DIFF
--- a/backend/db/config.py
+++ b/backend/db/config.py
@@ -6,6 +6,8 @@ from backend.config import CONFIG, CONFIG_LOCAL
 
 def invert_list_dict(d1: Dict[str, List[str]]) -> Dict[str, List[str]]:
     """Invert a dictionary with lists as values"""
+    if any(not isinstance(value, list) for value in d1.values()):
+        raise ValueError('Dictionary values must be lists')
     d2 = {}
     for key, values in d1.items():
         for value in values:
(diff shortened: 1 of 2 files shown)


backend/db/config.py
```diff
@@ -6,6 +6,8 @@
 
 def invert_list_dict(d1: Dict[str, List[str]]) -> Dict[str, List[str]]:
     """Invert a dictionary with lists as values"""
+    if any(not isinstance(value, list) for value in d1.values()):
+        raise ValueError('Dictionary values must be lists')
```
joeflack4, line 10: FYI: catches invalid config type. Honestly I should support 'string' instead of 'list' if dependency is a single table though.

backend/db/config.py
```diff
@@ -93,8 +95,12 @@ def get_pg_connect_url(local=False):
 ]
 # STANDALONE_TABLES: Not derived from any other table, nor used to derive any other table/view. Used for QC testing.
 STANDALONE_TABLES = [
+    'concept_set_json',
     'junk',
     'relationship',
+    'session_concept',
+    'sessions',
+    'test',
```
joeflack4, line 103: New standalone tables in `n3c`:
    'concept_set_json',
    'session_concept',
    'sessions',
    'test',

backend/db/config.py
```diff
@@ -103,7 +109,7 @@ def get_pg_connect_url(local=False):
 # - Last updated: 2024/04/07
 # todo: are the 'module names' part of the DDL file names the same as the keys in this dictionary? If so, write that
 #  down here as a comment. Could be useful for dynamically selecting the DDL files.
-DERIVED_TABLE_DEPENDENCY_MAP = {
+DERIVED_TABLE_DEPENDENCY_MAP: Dict[str, List[str]] = {
```
joeflack4, line 112: New relationships:
- concept_depth
- root_concepts

Vocab refresh wil now update these

joeflack4: FYI, when this gets updated, in order for the vocab refresh to know about these changes, I do a `rebase` of the `shadow-develop` branch on: 
- https://github.com/shadow-bids/termhub/

The main reason I do a `rebase` instead of a merge is so that there is a single commit difference between these branches, and I like it to show up at the top of the history. That commit deactivates every GH action, except for vocab/counts, which it activates.

PR #793, opened 2024-05-17
Bug: Vocab refresh fail: `concept_depth`

joeflack4:
## Changes
    Bug fix: Vocab refresh error
    - Bug fix: Fixed error in which temporary tables were being treated as permanent, which caused error when it tried to manipulate a table that was no longer there.
    - Update: Enabled extensions for all schema, so they are available for testing and future proof situations where extension is needed.
    - Bug fix: Removed temp tables from config, which was causing them to update when they should not (caused error).

    General
    - Bug fixes: Jinja updates to SQL DDL, fixing various possible unforeseen current and/or future issues.
    - Delete: Some old no-op temp code

### DIFF
--- a/backend/db/config.py
+++ b/backend/db/config.py
@@ -124,8 +124,7 @@ DERIVED_TABLE_DEPENDENCY_MAP: Dict[str, List[str]] = {
     ],
     'codeset_counts': ['members_items_summary'],
     'codeset_ids_by_concept_id': ['cset_members_items'],
-    'concept_ancestor_plus': ['concept_ancestor', 'concepts_with_counts', 'concept_depth'],
-    'concept_depth': ['root_concepts', 'concept_ancestor'],
+    'concept_ancestor_plus': ['concept_ancestor', 'concepts_with_counts'],
     'concept_graph': ['concept_ancestor'],
     'concept_ids_by_codeset_id': ['cset_members_items'],
     'concept_relationship_plus': ['concept_relationship', 'concepts_with_counts'],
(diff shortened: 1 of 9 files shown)


backend/db/config.py
```diff
@@ -124,16 +124,14 @@ def get_pg_connect_url(local=False):
     ],
     'codeset_counts': ['members_items_summary'],
     'codeset_ids_by_concept_id': ['cset_members_items'],
-    'concept_ancestor_plus': ['concept_ancestor', 'concepts_with_counts', 'concept_depth'],
-    'concept_depth': ['root_concepts', 'concept_ancestor'],
+    'concept_ancestor_plus': ['concept_ancestor', 'concepts_with_counts'],
```
joeflack4, line 127: ## Delete: `concept_depth` and `root_concepts` from dependency config
These are temp tables.

Deleting these from config not only reduces unnecessary operations during the refresh, but fixes an error.

backend/db/initialize.py
```diff
@@ -82,12 +82,14 @@
...
-# If, when running initialize() there is an error because can't create the extension, try step (1) and enable in Azure.
+# 2. Run SQL: `CREATE EXTENSION <EXTENSION> WITH SCHEMA pg_catalog`;
+# - if, when running initialize() there is an error because can't create the extension, try step (1) and enable in Azure.
+# - `WITH SCHEMA pg_catalog`: This enables the extension for all schemas, which is good for future proofing. Currently
+# it will mainly be used in the main schema, `n3c`, but might also be used in the test schema, `test_n3c`.
 DDL_EXTENSIONS = """
-    CREATE EXTENSION PG_TRGM;
-    CREATE EXTENSION BTREE_GIN;
-    CREATE EXTENSION BTREE_GIST;"""
+    CREATE EXTENSION PG_TRGM WITH SCHEMA pg_catalog;
```
joeflack4, line 90: Enabled extensions on all schema. Useful if need in test schema, or at least don't want test schema to error out, or if we need in other schema.

joeflack4: FYI as I mentioned on Teams, I also reinstalled the extensions. Didn't fix my error though.

PR #808, opened 2024-06-02
Bugfix: Backend tests (2024-06-02)

joeflack4:
## Changes
    Bugfix: Backend tests
    - Bug fixes: Failing test_resolve_fetch_failures_excess_items() and FetchAuditTestRunner() tearDownClass assertion.
    - Updates: Minor changes including param order, comments, and removing a print statement.

### DIFF
--- a/backend/db/resolve_fetch_failures_excess_items.py
+++ b/backend/db/resolve_fetch_failures_excess_items.py
@@ -18,15 +18,16 @@ from enclave_wrangler.utils import was_file_modified_within_threshold
 
 DESC = "Resolve failures due to too many expression items or members when fetching data from the Enclave's objects API."
 
-def resolve_fetch_failures_excess_items(schema=SCHEMA, use_local_db=False, cached_dataset_threshold_hours=0):
+def resolve_fetch_failures_excess_items(schema=SCHEMA, cached_dataset_threshold_hours=0, use_local_db=False):
     """Resolve failures due to too many expression items or members when fetching data from the Enclave's objects API.
-    cached_dataset_threshold_hours: Threshold, in hours, until cached datasets considered invalid and need to be
+
+    Uses the datasets API instead, to get around object API limits.
+
+    param cached_dataset_threshold_hours: Threshold, in hours, until cached datasets considered invalid and need to be
     re-downloaded.
 
     todo: rename to _items_members (it fetches excess caes of both)
     """
-    print('Resolve failures due to too many expression items or members when fetching data from the Enclave\'s '
-          'objects API by using datasets API instead.')
     # Vars
     datasets = ['concept_set_version_item', 'concept_set_members']
     failure_dataset_map = {
@@ -41,6 +42,7 @@ def resolve_fetch_failures_excess_items(schema=SCHEMA, use_local_db=False, cache
     if not failures:
         return
 
+    # Determine which datasets are needed to resolve failures
     datasets_needed: Set[str] = set()
     failures_by_dataset: Dict[str, List[Dict]] = {}
     for failure in failures:
@@ -54,7 +56,8 @@ def resolve_fetch_failures_excess_items(schema=SCHEMA, use_local_db=False, cache
     # - Determine if cache OK
     for dataset in list(datasets_needed):
         path = dataset_path_map[dataset]
-        if cached_dataset_threshold_hours and was_file_modified_within_threshold(path, cached_dataset_threshold_hours):
+        if cached_dataset_threshold_hours and os.path.exists(path) and \
+            was_file_modified_within_threshold(path, cached_dataset_threshold_hours):
             datasets_needed.remove(dataset)
     # - Download
     if datasets_needed:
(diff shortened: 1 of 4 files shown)


backend/db/resolve_fetch_failures_excess_items.py
```diff
@@ -78,7 +79,7 @@ def resolve_fetch_failures_excess_items(schema=SCHEMA, use_local_db=False, cache
                     if rows:
                         insert_from_dicts(con, dataset, rows)
                         solved_failures.append(failure)
-            refresh_derived_tables(con)
+            refresh_derived_tables(con, schema=schema)
```
joeflack4, line 82: Bug fix: Part of `resolve_fetch_failures_excess_items()` did not include `schema` param, so wasn't running on test schema if requested.

backend/db/resolve_fetch_failures_excess_items.py
```diff
@@ -54,7 +56,8 @@ def resolve_fetch_failures_excess_items(schema=SCHEMA, use_local_db=False, cache
     # - Determine if cache OK
     for dataset in list(datasets_needed):
         path = dataset_path_map[dataset]
-        if cached_dataset_threshold_hours and was_file_modified_within_threshold(path, cached_dataset_threshold_hours):
+        if cached_dataset_threshold_hours and os.path.exists(path) and \
```
joeflack4, line 59: Bug fix: If allowing use of cached dataset and the dataset had never been downloaded before (e.g. GH action), was erroring.

backend/db/utils.py
```diff
@@ -142,7 +142,7 @@ def refresh_derived_tables_exec(
      entry will appear further down in the list."""
     temp_table_suffix = '_new'
     ddl_modules_queue = derived_tables_queue
-    views = [x for x in list_views() if x in ddl_modules_queue]
+    views = [x for x in list_views(schema=schema) if x in ddl_modules_queue]
```
joeflack4, line 145: Bug fix: Refreshes on `test_n3c` could error out because `list_views()` was not detecting views.

The reason that view `concept_set_member_plus` exists in `test_n3c` but not `n3c` is because it still exists in the DDL, which is no longer re-run for `n3c`, I believe, whereas `test_n3c` gets regenerated each time the tests run.

backend/db/utils.py
```diff
@@ -182,6 +182,9 @@ def refresh_derived_tables_exec(
 # todo: move this somewhere else, possibly load.py or db_refresh.py
 # todo: what to do if this process fails? any way to roll back? should we?
 # todo: currently has no way of passing 'local' down to db status var funcs
+# todo: last_derived_refresh_request and last_derived_refresh_exited: these should ideally be schema specific. That way,
```
joeflack4, line 185: Minor optimization todo: allow `test_n3c` and `n3c` refreshes to run simultaneously.

test/test_backend/db/test_utils.py
```diff
@@ -18,6 +18,8 @@
     select_failed_fetches, sql_query
 
 
+# todo: add datetime to setUp and tearDown: It might be possible, despite failsafes being in place to prevent refreshes
```
joeflack4, line 21: Minor `todo`: It's possible for test assertion to fail if live data was inserted during the course of tests running, but I think given the nature of this test, and due to only 1 refresh being able to run at a time, that's a very low probability.

test/test_backend/db/test_utils.py
```diff
@@ -33,10 +35,10 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         """tearDown"""
-        run_sql(cls.con, f"DELETE FROM fetch_audit WHERE comment = 'Unit testing.';")
+        run_sql(cls.con, f"DELETE FROM fetch_audit WHERE comment LIKE 'Unit testing.%';")
```
joeflack4, line 38: Edge case fix: It might be possible for this not to delete rows as intended, because `comment` field can get updated (e.g. by adding results of fetch resolution).

test/test_backend/test_resolve_fetch_failures.py
```diff
@@ -45,18 +47,25 @@ class TestBackendResolveFetchFailures(FetchAuditTestRunner):
...
         if self.run_live_fetch:
             run_sql(self.con, f"DELETE FROM fetch_audit WHERE primary_key = '{pk}' AND comment = 'Unit testing.';")
             fetch_cset_and_member_objects(codeset_ids=[pk])
-        status1 = query()
-        resolve_fetch_failures_excess_items()
-        status2 = query()
+        status1 = self._failure_status_query(pk)
+        # todo: Ideally should be able to pass specific failures. What if actual failures exist that aren't test cases?
+        #  the it will run those as well.
+        resolve_fetch_failures_excess_items(TEST_SCHEMA, 24)
```
joeflack4, line 67: ## The main bug fixes (workaround solution)
I'm not actually sure what the source of these original errors were. One was happening during a `resolve_fetch_failures_excess_items()` assertion. It was checking on the status of a temporary mock fetch failure, but for some reason it got deleted early.

There was another but in `FetchAuditTestRunner`, where the number of rows was different at the end of the test. I inserted a todo that might address the issue (data being inserted by a live refresh while tests running), but I think this might be a rare case, so I don't know if it's a huge priority. I could be wrong about how rare that is, though, in which case I'll fix this sooner. Other than that though I'm not sure what could be causing the issue.  But the fact that it happened simultaneously with the `resolve_fetch_failures_excess_items()` issue above makes me think that they are related.

Workaround solution
I say workaround, but it's really a side effect of changing the tests to be set up in the way that they should have been. They were running off of `n3c`. The test was actually triggering a refresh of that schema, not `test_n3c`. After fixing this, the errors went away.

Issue #807, opened 2024-06-02
`test_subgraph()`


joeflack4:

joeflack4:
More notes in Workflowy; will discuss with Siggie.

Issue #784, opened 2024-05-06
Bug fixes & tests: missing in-betweens


joeflack4:
## Overview
Siggie wrote:
> I discovered a bug in get_missing_in_between_nodes
> We need to figure out why our existing tests didn't catch it and fix them.
> ```python
> from backend.routes.graph import get_missing_in_between_nodes, REL_GRAPH
> x = get_missing_in_between_nodes(REL_GRAPH, [4146581, 317009, 257581], verbose=True)
> 
> x = get_missing_in_between_nodes(REL_GRAPH, [4146581, 317009], verbose=True)
> ```
> Two bugs, actually.
> 317009 is an ancestor of 4146581, but this call returns nothing in between. The problem there is (I think) that 317009 is considered a leaf because nothing is connected to it in this 2-item subgraph. Adding 257581 makes it so 317009 is seen as a root. > But it still doesn't work. I haven't figured out why yet.
> Crap... this algorithm is tricky. I've spent a couple hours this morning on it and am kind of baffled. It needs better documentation and I probably need help debugging it.

joeflack4:
I was thinking this would be a good opportunity for TDD, so I made a single issue for both, but Siggie completed the bug fixes already and I don't think they saw my message, so since all that remains is the test, I renamed the issue.

joeflack4:
Note to self: I have a lot of notes about this in Workflowy.

PR #810, opened 2024-06-16
Graph tests

joeflack4:
## Updates
    Graph tests
    - Update: test_subgraph(): Moved from test_backend.py to test_graph.py
    - Update test_graph.py: Activated tests
    - Delete: Deprecated get_connected_subgraph(). Has moved to concept_graph()
    - Update: Needed to refactor these tests to match the codebase changes. Still much broken, unwritten assertions; WIP.

## Priority
Siggie thought it might be quick to reactivate `test_subgraph()`. I'm also activating other tests. If they break and they're not quick to fix, I'm going to set them to skip, and make an issue to do them later, as other tests / issues may have greater priority.


test/test_backend/routes/test_graph.py
joeflack4, line 1: ## Reactivated tests
By renaming the file, methods, class, and assertions.

test/test_backend/test_backend.py
```diff
@@ -339,134 +339,3 @@ def test_get_cset_members_items__cols(self):
             {'concept_id': 4091006, 'standard_concept': 'S', 'vocabulary_id': 'SNOMED'},
             {'concept_id': 4052321, 'standard_concept': 'S', 'vocabulary_id': 'SNOMED'}].sort(key=key)
         self.assertEquals(csmi, expected)
-
-    # TODO: test is failing. fix
-    #  - The cause of the err is that subgraph() is a different function now. it used to call connected_subgraph_from_nodes(), which no longer exists.
-    #    Is this test still valid? Is there any func that takes a list of concept IDs and returns their parentage tuples?
-    @unittest.skip("May need to remove this test. I think the code it was testing got removed. Skipping failing/erroring test temporarily.")
-    def test_subgraph(self):
```
joeflack4, line 347: ## Moved `test_subgraph()`
To `test_graph.py`. Still doesn't work. That's a separate issue.

test/test_backend/routes/test_graph.py
```diff
@@ -0,0 +1,390 @@
...
+            'orphans_not_in_graph': sorted(list(orphans_not_in_graph)),
+            'hidden': normalized_hidden,
+        }
+        n_records = sum([len(v) for k, v in results.items()])
+        indent = None if n_records > minify_threshold else 4
+        with open(path, 'w') as f:
+            f.write(json.dumps(results, indent=indent))
+
+    # todo: this is not currently used. It needs modification after changes to get_connected_subgraph()
+    async def _create_inputs__get_connected_subgraph__needs_repair(self):
```
joeflack4, line 115: ## Drop `async`
siggie async def don't run in normal suite, else if run manually they just pass without doing anything
Apparently with `async`, these don't run in the normal test suite when called.

I think if you still need to run these `async`, we should split most of the test into a separate function. Hopefully that way (i) you can call it when you need, as async, and (ii) it will run via the test suite.

joeflack4: obsolete. i have since fixed async to work with tests

test/test_backend/routes/test_graph.py
```diff
@@ -0,0 +1,402 @@
...
+"""
+TEST_CASES_HOLD = """
+testType,testName,codeset_ids,timeoutSeconds
+"""
+TEST_CASES = TEST_CASES_FAST
+EXTRA_CONCEPT_IDS = []
+HIDE_VOCABS = ['RxNorm Extension']
+
+
+class AsyncTests(unittest.TestCase):
```
joeflack4, line 54: ## Got `async` tests to work
They wouldn't run with the normal test suite. I had to write this wrapper class. [ChatGPT helped](https://chatgpt.com/c/d9859ecb-c919-44bd-b8f5-8e8c9f09e2f6).

test/test_backend/routes/test_graph.py
joeflack4: @Sigfried Priority questions:
You gave me an idea of what's needed for #807. Should be "quick" hopefully, but we're not using that code at all.

Meanwhile, #784 is passing with one test case, but you wanted me to add more cases.

Finally, #811 has 2 test methods with different scenarios, and they are failing / assertions need updated.

Would you say that #784 is still a higher priority than #811?

PR #812, opened 2024-06-21
Bugfix: airium dependency

joeflack4:


requirements.txt
```diff
@@ -1,6 +1,6 @@
 aiohttp==3.8.4
 aiosignal==1.3.1
-airium==0.2.5
+airium==0.2.6
```
joeflack4, line 3: Addressed issue where a new version of pip did not allow for our fixed version of this dependency to be installed. Incremented the dependency, fixing the problem.

Issue #819, opened 2024-06-24
Tests: Have LLMs create many for us


joeflack4:
## Overview
Apparently GPT 4.0 (IDK about 3.5 or 4o, or other LLMs) is really good at creating unit tests.
We can't just copy/paste the whole codebase into the prompt yet, but we could do single files at a time.
I think this could be a great idea to quickly (i) stress test the codebase (it can find good edge cases), (ii) future proof errors as a result of changes.

PR #828, opened 2024-07-06
Tests: Have LLMs create many for us - objects_api - 1st go

joeflack4:
Partially addresses:
- #819

## Changes
Unit tests: new battery of tests generated from AI
Using generative AI, e.g. Claude, to look at individual files and come up with some test case scaffolding for us, then tweaking to fit our codebase, expected API results, etc.
- Update: test_objects_api.py with such tests

General
- Update: link_types(): It now just calls get_link_types(), and it is now deprecated. Functions were doing the same thing, only link_types() had less options.
- Add: Some todo comments
- Update: Some misc minor refactoring
- Update: Changed a get_object_links() typedef


test/test_enclave_wrangler/input/test_get_object_types/object_types.json
joeflack4, line 1: Committed the return of this func as an input.

test/test_enclave_wrangler/test_objects_api.py
```diff
@@ -211,3 +233,206 @@ def test_get_concept_set_version_members(self):
         obj_id = 563193300
         data = get_concept_set_version_members(obj_id, return_detail='full')
         self.assertGreater(len(data), 0)
+
+
+# TODO: new tests by claude to try
+# from pprint import pprint as pp
+# todo: #1: Wherever I put '#1', consider removing the mock and call the actual function
+class TestObjectsApiMocks(DbRefreshTestWrapper):
```
joeflack4, line 241: New class that [Claude made](https://claude.ai/chat/7f64ea42-7ab2-43c6-9a86-382bdafc26bf) to do mock API tests.

test/test_enclave_wrangler/test_objects_api.py
```diff
@@ -211,3 +233,206 @@ def test_get_concept_set_version_members(self):
...
+        # test_get_object_types()
+        get_object_types: os.path.join(TEST_INPUT_DIR, 'test_get_object_types', 'object_types.json')
+    }
+
+    def setUp(self):
+        """Set up mock connection"""
+        self.mock_connection = MagicMock()
+
+    @classmethod
+    def setUpClass(cls):
```
joeflack4, line 256: ## Set up: Storing returns of functions we are mocking
At the beginning of any tests using this class, sees if the function being mocked has a cached version of its outputs saved. If not, it fetches and saves in the test input dir.

test/test_enclave_wrangler/test_objects_api.py
```diff
@@ -211,3 +233,206 @@ def test_get_concept_set_version_members(self):
...
+        """Read cached return from file"""
+        path = self.func_cached_return_map[func]
+        with open(path, 'r') as f:
+            return json.load(f)
+
+    # TODO: dryify this to re-use?
+    #  - one difficulty is {'data': response} and other optional small mutations various funcs implement.
+    @patch('enclave_wrangler.objects_api.enclave_get')  # todo #1
+    def test_get_object_types(self, mock_enclave_get):
+        """Test get_object_types()
```
joeflack4, line 274: ## Test `get_object_types()`
The first test. It's kind of mundane. But basically I make sure we get back at least these endpoints:
`'OMOPConcept', 'OMOPConceptSetContainer', 'OMOPConceptSet', 'OmopConceptSetVersionItem'`

However this test is just a mock. I think it's good to have mock tests, just to make sure that no essential basic logic around API calls gets botched. However, it would indeed be good to call the actual enclave API itself.

enclave_wrangler/objects_api.py
```diff
@@ -182,33 +185,10 @@ def get_ontologies() -> Union[List, Dict]:
     response_json = response.json()
     return response_json
 
-
+@deprecated(details="Use get_link_types() instead")
 def link_types() -> List[Dict]:
```
joeflack4, line 189: ## Deprecated `link_types()`
I checked and this was doing the same thing (only w/ less options) as `get_link_types()`.

Issue #398, opened 2023-06-15
DB refresh: Updated objects (i.e. metadata)


joeflack4:
## Overview
Currently the DB refresh is set up so that any new created objects created since a certain time will be fetched. However, data about any updates / archivals (I don't think) / deletions of objects will not be fetched.

Since these objects are mostly immutable, I think this only applies to metadata.

## Involves the following objects
- concept set containers
- concept set versions
- member items
- expression items
- other objects (#341)

Sigfried:
Solve by changing fetch since queries to compare to modifiedAt instead of createdAt

joeflack4:
Thanks Siggie. That is the most important thing. Other logic will need to be changed too. If I change that one line and do nothing else, it will try to insert data as if it is a new record rather than doing an update. Shouldn't require that many changes in the logic, I don't think.

Sigfried:
@joeflack4, where does this stand? Is it done?

And can you confirm, I think we decided to continue checking createdAt in addition to modifiedAt. Is that happening?

joeflack4:
@Sigfried No, I haven't started on this yet. This is the only/most critical DB refresh issue after fixing the vocab/counts refresh GH action running out of memory, and I aim to do this immediately after that.

We will need to continue checking both `createdAt` and `modifiedAt`. Main reason for this is that the latter is often NULL. Ideally it would be >= `createdAt`.

PR #836, opened 2024-07-13
Update finalized draft metadata

### LINKED ISSUE #520: New Feature: Drafts

joeflack4:
resolves #520

## Changes
Update finalized draft titles
When a draft is finalized, make sure to remove [DRAFT] from the title, and make sure to include the correct version number in the version title.
- Update: resolve_fetch_failures_0_members.py: To do this whenever drafts get finalized.
- Add: update_finalized_draft_metadata()
- Add: update_cset_metadata_from_objs()
- Add: test_update_cset_metadata_from_objs()
- Add: update_from_dicts()
- Add: update_objects_in_db()
- Add: clean_improper_finalized_draft_metadata()

General
- Update: test_objects_api.py: Made a single connection for test class and re-used.
- Update: requirements-unlocked.txt: Some comments
- Update: backend/db/utils.py: Codestyle fixes (typing, comments, etc)
- Update: convert_row(): Added param to keep fields even if they're missing from the 'target'. This helps in situations where you are converting a row from object to table model or vice versa, but it is already in the correct model format. In those cases, we don't want the object to be changed.

## Related
- #398 
- #520


backend/db/initialize.py
```diff
@@ -155,6 +155,14 @@ def initialize(
 
         make_derived_tables_and_more(con, schema, local=local)  # , start_step=30)
 
+        # Make update audit tables
+        # - these tables are to store history of updates made to 'core' tables; when they were made, not by who
+        audit_update_tables = ['code_sets']
```
joeflack4, line 160: ## Audit tables for updates & deletes
I set up a way to do this, so we can see the history of updates and possibly undo changes.

Added to `initialize()`.

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -244,13 +245,16 @@ def resolve_fetch_failures_0_members(
...
+            [x['properties'] for x in success_cases['OMOPConceptSet'] if x['properties']['codesetId'] in finalized_draft_ids]
 
         # Update DB & report success
         if success_cset_ids:
             try:
                 with get_db_connection(schema=schema, local=use_local_db) as con:
-                    csets_and_members_to_db(con, success_cases, schema)
-                    refresh_derived_tables(con)
+                    update_cset_metadata_from_objs(finalized_drafts, con)
+                    csets_and_members_to_db(con, success_cases, ['OmopConceptSetVersionItem', 'OMOPConcept'], schema)
```
joeflack4, line 257: ## Finalized drafts now update metadata

backend/db/utils.py
```diff
@@ -565,6 +564,45 @@ def get_cset_members_items_rows(
...
+    fields = []
+    for row in rows:
+        fields.extend(row.keys())
+        for k in row.keys():
+            fields.append(k)
+    fields = set(fields)
+    return [{field: row.get(field, None) for field in fields} for row in rows]
+
+
+def update_from_dicts(con: Connection, table: str, rows: List[Dict]):
```
joeflack4, line 578: ## Update function (table objects)
Pass table name and a list of table objects.

Will update any rows matching the primary keys in the objects.

enclave_wrangler/objects_api.py
```diff
@@ -648,6 +658,16 @@ def fetch_cset_expression_item(object_id: int, retain_properties_nesting=False)
     return fetch_object_by_id('OmopConceptSetVersionItem', object_id, 'itemId', retain_properties_nesting)
 
 
+def update_objects_in_db(
```
joeflack4, line 661: ## Update function (enclave objects)
Converts them into table objects and then kicks off updates.

enclave_wrangler/objects_api.py
```diff
@@ -1024,6 +1045,80 @@ def get_projects(verbose=False) -> List[Dict]:
...
+    #   - did my modification below to use PG_DATATYPES_BY_GROUP work? run it
+    return [
+        cset | {
+            f: 0 if obj_data_types[f] in PG_DATATYPES_BY_GROUP['numeric'] else None
+            for f in obj_fields.difference(cset.keys())}
+        for cset in cset_objs
+    ]
+
+
+def update_cset_metadata_from_objs(
```
joeflack4, line 1081: ## `update_cset_metadata_from_objs()`
Wrapper for:
1. Updates the 'audit' table
2. Kicks off the updates

enclave_wrangler/objects_api.py
```diff
@@ -1024,6 +1045,80 @@ def get_projects(verbose=False) -> List[Dict]:
...
+    for cset in cset_objs:
+        cset_id = str(cset['codesetId'])
+        run_sql(con, audit_query.replace('cset_id', cset_id))  # backup to audit table: original
+        update_objects_in_db(con, 'OMOPConceptSet', [cset])
+        run_sql(con, audit_query.replace('cset_id', cset_id))  # backup to audit table: updated
+    if not con:
+        conn.close()
+
+
+def clean_improper_finalized_draft_metadata():
```
joeflack4, line 1117: ## `clean_improper_finalized_draft_metadata()`
In the `objects_api` CLI.

Identify finalized drafts with improper metadata and update their metadata, title and all.

test/test_enclave_wrangler/test_objects_api.py
```diff
@@ -280,6 +276,93 @@ def test_get_csets_over_threshold(self):
...
+        cset_id_str: str = ', '.join([str(x) for x in cset_ids]) if cset_ids else ''
+        # TODO: test_n3c shouldn't be necessary, but GH action is failing.
+        query: str = f"""SELECT * FROM code_sets WHERE codeset_id IN ({cset_id_str});""" if cset_ids \
+            else f"""SELECT * FROM code_sets LIMIT {limit};"""
+        csets: Dict[int, Dict] = {x['codeset_id']: dict(x) for x in sql_query(cls.con, query)}
+        if convert_to_enclave_object:
+            csets = {k: convert_row('code_sets', 'OMOPConceptSet', v) for k, v in csets.items()}
+        return csets
+
+    def test_update_cset_metadata_from_objs(self):
```
joeflack4, line 295: ## `test_update_cset_metadata_from_objs()`
1. Does arbitrary mutations on all fields, and updates row in DB.
2. Fetches from enclave and updates again.

enclave_wrangler/objects_api.py
```diff
@@ -1024,6 +1045,118 @@ def get_projects(verbose=False) -> List[Dict]:
...
+        for cset in csets_db:
+            csets_enclave.append(fetch_cset_version(cset['codeset_id']))
+            # todo: I don't think i need this
+            # cset['properties']['isDraft'] = True
+            # cset['properties']['title'] = cset['properties']['title'].replace('[DRAFT]', '').strip()
+        # TODO: Filter only by those which changed
+
+        csets_enclave2 = cset_objs_set_missing_fields_to_null(csets_enclave)
+        csets_enclave_table_objs: List[Dict] = [convert_row('OMOPConceptSet', 'code_sets', x) for x in csets_enclave2]
+        cset_table_keys_not_in_objs: Set[str] = set(
```
joeflack4, line 1139: ## Missing fields from enclave objects?
@Sigfried FYI, these field are in `code_sets` but do not come back when we query the enclave. Do you think it's because these 700 csets I fetched just didn't have any data on these fields? Or are these not available via the enclave API? If not, then how did these get in our `code_sets` table? From datasets API?

- `atlas_json`
- `authoritative_source`
- `comments`
- `date_submitted_zenodo`
- `first_author`
- `has_review`
- `issues`
- `reviewed_by`
- `status`

backend/db/utils.py
```diff
@@ -462,15 +462,59 @@ def database_exists(con: Connection, db_name: str) -> bool:
...
+    # Make replacements
+    query2 = query
+    for fld in numeric_fields_with_none:
+        query2 = query2.replace(f'"{fld}" = v.{fld}',
+            f'"{fld}" = CASE WHEN v.{fld}::text = \'None\' THEN NULL ELSE v.{fld}::double precision END')
+    return query2
+
+
+# todo: schema: Ideally, I should make sure that any usages of this function are passing `schema`.
+def run_sql_update(
```
joeflack4, line 498: ## `run_sql_update()`
If you get an error related to NULL while doing an update query, use this in combination with the `handle_none_as_null_on_table` param.
Basically, I want to ensure that if the user deletes or submits empty text / null for a metadata field, we can support that. Normally it is not an issue, except for numeric fields. 

This was a pain in the ass to fix. I tried different things, like `sqlalchemy.null()` and 
`sqlalchemy.Float().bind_processor(dialect=con.dialect)(None)`, but fixed another way. Now I question whether there's ever a situation we really _need_ this support. I'm looked at the numeric fields, and I'm not sure if they can ever go from a non-null to a null value. Oh well.

enclave_wrangler/models.py
```diff
@@ -383,16 +383,24 @@ def convert_rows(source: str, target: str, rows: List[Dict]) -> List[Dict]:
     return out
 
 
-def convert_row(source: str, target: str, row: Dict, skip_missing_fields=True) -> Dict:
-    """Convert row from source to target"""
+# todo: automatically determine the source and target rowtypes based on the fields in the row
+# todo: consider, if running multiple times on row, do we want it to error the second time, since the row has already
+#  been converted, or do we want it to do nothing and return the same obj?
+def convert_row(source: str, target: str, row: Dict, skip_missing_fields=True, keep_missing_fields=False) -> Dict:
```
joeflack4, line 389: ## `convert_row()`: Added `keep_missing_fields`
If True, keep fields that are not in the target model. This is useful when the source is actually the same object type as the target. If True, will result in the returned object/row being the same as the `row` that was passed.

enclave_wrangler/objects_api.py
```diff
@@ -1024,6 +1044,157 @@ def get_projects(verbose=False) -> List[Dict]:
...
+                db_value = db_obj.get(key)
+                enclave_value = enclave_obj.get(key)
+                if db_value != enclave_value:
+                    differences[key] = (db_value, enclave_value)
+            if differences:
+                report[codeset_id] = differences
+    return report
+
+
+def clean_improper_finalized_draft_metadata(check_even_if_still_draft=False, verbose=False, codeset_ids: List[int]=None):
```
joeflack4, line 1139: ## Fixed `[DRAFT]` in finalized cset titles
Checked any csets where title `LIKE` `[DRAFT]`, and made sure their metadata was up-to-date. This resulted many fixes where we did not update the cset when the draft was finalized. There were also some cases of csets that were still drafts but that have had metadata changes too.

PR #846, opened 2024-07-28
Bug fix: NULL concept name

joeflack4:
## Changes
Bug fix: NULL concept name
- When adding concept_members to the DB, if there was no concept name, would cause an exception.


enclave_wrangler/objects_api.py
```diff
@@ -740,7 +740,7 @@ def concept_set_members__cset_rows_to_db(con: Connection, cset: Dict, members: L
         'concept_set_name': cset['conceptSetNameOMOP'],
         'is_most_recent_version': cset['isMostRecentVersion'],
         'version': cset['version'],
-        'concept_name': member['conceptName'],
+        'concept_name': member.get('conceptName', None),
```
joeflack4, line 743: ## The bug fix
I was surprised to see this. Apparently, out of ~7.5m concepts, only 41 of them don't have a concept name. All but 4 of them are from CPT4. About half are standard concepts. Only 1 is deprecated.

PR #847, opened 2024-07-29
GH Actions: Fewer installs

joeflack4:
## CHanges
    GH actions: Fewer installs
    - Deactivated install --upgrade of pip, wheel, and setuptools, as this probably decreases stability moreso than increases.

### DIFF
--- a/.github/workflows/backend_dev.yml
+++ b/.github/workflows/backend_dev.yml
@@ -49,9 +49,10 @@ jobs:
           git lfs pull
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install --upgrade wheel
-          pip install --upgrade setuptools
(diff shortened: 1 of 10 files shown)


.github/workflows/backend_dev.yml
```diff
@@ -49,9 +49,10 @@ jobs:
           git lfs pull
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install --upgrade wheel
-          pip install --upgrade setuptools
+          # Reactivate install upgrades of pip, wheel, &/or setuptools if needed
+          # python -m pip install --upgrade pip
+          # pip install --upgrade wheel
+          # pip install --upgrade setuptools
```
joeflack4, line 55: Deactivated

PR #853, opened 2024-08-25
DB refresh: by `codeset_id`

### LINKED ISSUE #832: DB refresh: by `codeset_id`
## Overview
`fetch_cset_and_member_objects()` has a `codeset_ids` param. So we can at least grab data from the API for specific IDs. But the there is no param to actually request that specific IDs be imported / updated into the DB, surprisingly. Add this feature to `refresh.py`.

## Implementation
Should be pretty easy. Just add a function param, a CLI arg to match, and then pass the IDs down a couple functions into `fetch_cset_and_member_objects()`.

joeflack4:
resolves #832

DB Refresh: Support cset-ids
- Bug fix: Was not setting 'since' to None if cset-ids was present.
- Bug fix: Was passing cset-ids to secondary refresh (if a user requested a refresh from the web app during that refresh). That isn't supported by the UI.
- Update: Some comments and other minor things.
- Update: Now supported by the GitHub actions

### DIFF
--- a/.github/workflows/refresh_db.yml
+++ b/.github/workflows/refresh_db.yml
@@ -15,6 +15,11 @@ on:
         description: 'Optional: YYYY-MM-DD or YYYY-MM-DDTHH:MM:SS.SSSSSS+HH:MM'
         required: false
         type: string
+      cset-ids:
+        # Space-delimited list of integers representing codeset IDs
+        description: 'Optional: Space-delimited list of integers representing codeset IDs'
+        required: false
+        type: string
   repository_dispatch:
     types: [refresh-db]
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/refresh_db.yml
```diff
@@ -15,6 +15,11 @@ on:
         description: 'Optional: YYYY-MM-DD or YYYY-MM-DDTHH:MM:SS.SSSSSS+HH:MM'
         required: false
         type: string
+      cset-ids:
```
joeflack4, line 18: ## Github actions: now supported there

backend/db/refresh.py
```diff
@@ -124,6 +128,7 @@ def refresh_db(
     try:
         with get_db_connection(local=local) as con:
             since: str = _calc_refresh_datetime(con, since, buffer_hours, force_non_contiguity)
+            since = since if not cset_ids else None
```
joeflack4, line 131: Bug fix: Was not setting 'since' to None if cset-ids was present.

If both are passed, then a `NotImplementedError` would've gotten raised further down.

backend/db/refresh.py
```diff
@@ -156,7 +161,7 @@ def refresh_db(
     if check_db_status_var('new_request_while_refreshing'):
         print('INFO: New refresh request detected while refresh was running. Starting a new refresh.')
         delete_db_status_var('new_request_while_refreshing')
-        refresh_db(None,cset_ids,  use_local_db, schema, force_non_contiguity)
+        refresh_db(None, None, use_local_db, schema, force_non_contiguity)
```
joeflack4, line 164: Bug fix: Was passing cset-ids to secondary refresh (if a user requested a refresh from the web app during that refresh). That isn't supported by the UI.

joeflack4:
I had implemented some of this, but there were some minor bugs, and I gave it GitHub action UI support.

PR #854, opened 2024-08-26
Pydantic bugfix

joeflack4:
- Fixed a bug with RowMapping not being recognized by Pydantic during FastAPI serialization process after having upgraded it from 1.0 to 2.0. This was upgraded recently after adding VSHub SDK, but probably in error, as it now suggests a 1.0 version that was simply higher than what we had before.


requirements.txt
```diff
@@ -131,8 +131,7 @@ pyasn1==0.4.8
 pyasn1-modules==0.2.8
 pycg==0.0.6
 pycparser==2.21
-pydantic==2.8.2
-pydantic_core==2.20.1
+pydantic==1.10.9
```
joeflack4, line 134: ## The bug fix
This might look familiar. @Sigfried you mentioned we had this issue before. It was with LegacyRow and not RowMapping, and it was over a year ago. And I do believe it happened when we tried to upgrade pydantic and/or sqlalchemy.

I upgraded this after installing the `vshub_sdk` from the enclave. It was having an error with our current version of pydantic. I must not have read the message correctly though, because it was asking for `>=1.10.9`, not `>=2.0`. We were on `1.10.8`.

In any case, this version of pydantic seems to work fine.

BTW when I tested this, I believe that I just checked to make sure the tests passed and that the app started normally, but I didn't check various routes / general workflows, so my face check wasn't thorough enough.

PR #855, opened 2024-08-27
Bug fix: Type defs

joeflack4:
Something must have changed with one of our packages, maybe Starlette, or FastAPI, or Pydantic. Something is resulting in 'fastapi.exceptions.ResponseValidationError' on at least the route 'concept-ids-by-codeset-id'. Corrected the typedef for that route, and changed some other typedefs that didn't cause me errors, but I wondered if they might be safer just in case the code changes.

### DIFF
--- a/.run/backend.run.xml
+++ b/.run/backend.run.xml
@@ -2,7 +2,8 @@
   <configuration default="false" name="backend" type="Python.FastAPI">
     <option name="additionalOptions" value="--port 8000" />
     <option name="file" value="$PROJECT_DIR$/backend/app.py" />
-    <module name="TermHub" />
+    <module name="ValueSet-Tools" />
+    <option name="ENV_FILES" value="" />
     <option name="INTERPRETER_OPTIONS" value="" />
     <option name="PARENT_ENVS" value="true" />
     <envs>
--- a/backend/routes/db.py
+++ b/backend/routes/db.py
@@ -137,13 +137,14 @@ def get_cset_members_items(
 
 
 @router.get("/get-cset-members-items")
-async def _get_cset_members_items(request: Request,
-                                  codeset_ids: str = None,
-                                  columns: Union[List[str], None] = Query(default=None),
-                                  column: Union[str, None] = Query(default=None),
-                                  return_with_keys: bool = True,
-                                  # extra_concept_ids: Union[int, None] = Query(default=None)
-                                  ) -> Union[List[int], List]:
+async def _get_cset_members_items(
+    request: Request,
+    codeset_ids: str = None,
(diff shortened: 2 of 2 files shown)


.run/backend.run.xml
```diff
@@ -2,7 +2,8 @@
   <configuration default="false" name="backend" type="Python.FastAPI">
     <option name="additionalOptions" value="--port 8000" />
     <option name="file" value="$PROJECT_DIR$/backend/app.py" />
-    <module name="TermHub" />
+    <module name="ValueSet-Tools" />
```
joeflack4, line 5: IDK why my run config is changing back to this. I kept resetting it but finally it got commit by mistake; oh well. If this persists I'll try to fix.

termhub-vocab
joeflack4, line 1: @Sigfried I don't know why this changed. the last commit was yours from Nov 30. But this affects the `relationship_graph` pickles... so I wonder if this could break anything if they are somehow outdated. Maybe you should regenerate them just in case?

backend/routes/db.py
```diff
@@ -535,7 +534,7 @@ def download_n3c_recommended():
 
 
 @router.get("/n3c-recommended-report", response_model=False)
-def n3c_recommended_report(as_json=False) -> Union[List[Row], StreamingResponse]:
+def n3c_recommended_report(as_json=False):  # -> Union[List[Row], StreamingResponse]
```
joeflack4, line 537: ## Cautiously pre-emptively removed some extra typedefs
Maybe I was a bit overzealous, but I felt like, who knows, maybe some things could break in the future if these type definitions are too specific, so I commented some out and simplified some others.

backend/routes/db.py
```diff
@@ -257,9 +259,8 @@ async def get_concept_ids_by_codeset_id_post(request: Request, codeset_ids: Unio
 @return_err_with_trace
 async def get_concept_ids_by_codeset_id(
     request: Request, codeset_ids: Union[List[str], None] = Query(...)
-) -> Dict[str, str]:
+) -> Dict[int, List[int]]:
```
joeflack4, line 262: ## The bug fix
The typedef was wrong for this function. But IDK why it's erroring now whereas it was not before. It doesn't seem like our `requirements.txt` has changed very much, but probably one or more packages got upgraded in the past few months which was causing it to now do these validations.

termhub-vocab
Sigfried: Oh. Huh. When vocab changes, we should be regenerating this. I wonder how many problems that was causing?

All we need to do is delete the file and next time backend starts, it will get regenerated. Doing that now, and I'll commit it when done.

PR #875, opened 2024-09-11
Bug: Refresh fails; Dropping `concept` table?

joeflack4:
Bug fix: Issue dropping a table during refresh w/ dependent tables & no CASCADE

### DIFF
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -67,12 +67,13 @@ def extract_keys_from_nested_dict(d: Dict[str, Dict]) -> List[str]:
     return ordered_unique_keys
 
 
-def get_dependent_tables_queue(independent_tables: List[str]) -> List[str]:
+def get_dependent_tables_queue(independent_tables: Union[List[str], str], _filter: str = None) -> List[str]:
     """From independent_tables, get a list of all tables that depend on those tables.
 
     :param independent_tables:  The tables from which you are trying to get an ordered list of dependent tables from.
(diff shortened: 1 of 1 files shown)


backend/db/utils.py
```diff
@@ -919,11 +934,19 @@ def reset_temp_refresh_tables(schema: str = SCHEMA, local=False, verbose=True):
             # Get all tables/views
             items: List[str] = func(con)
             # _old tables/views
+            #  - For each, drop the current one and replace w/ the backed up one. Drop and remake any dependent views.
             backed_up_items = [t for t in items if t.endswith('_old')]
             for item in backed_up_items:
                 item = item.replace('_old', '')
+                dependent_views: List[str] = get_dependent_tables_queue(item, 'views')
+                for view in dependent_views:
+                    run_sql(con, f'DROP VIEW IF EXISTS {schema}.{view};')  # alternative: drop table w/ CASCADE
```
joeflack4, line 943: ## The bug fix
I think that the vocab refresh failed because of our token issue, I'm not sure. But when a refresh fails, there is a "reset table state" procedure. It could not complete this procedure for the `concept` table because it had dependent view `cset_members_items_plus`.

I know we weren't using this before, and maybe we're not using it now. So rather than just solving this by just dropping that view permanently, I added code to drop and regenerate any views it needs to during the process.

Note that we do have a new view `all_csets_view`, so this update will probably be necessary to for that, too.

PR #877, opened 2024-09-11
Various DB config/refresh bug fixes

joeflack4:
Various DB config/refresh bug fixes
    - Bug fix: `cset_members_items_plus` sometimes being dropped.
    - Bug fix: "test_n3c.all_csets_view_new" does not exist. Test schema views were not being created correctly.
    - Bug fix: An error that was causing duplicate entries to be added to the `concept` table.
    - Bug fix: cannot drop table all_csets_old because other objects depend on it
    - Update: backend/db/config.py: Simplified by deleting unnecessary CORE_CSET_DEPENDENT_TABLES. This was used as the default value of a function, but that default was never being used. What's more, it is easily computed (that's what's currently being done). Keeping this variable meant more to understand/maintain for no real purpose.
    - Update: Docs for adding new view/table.


backend/db/ddl-12-cset_members_items_plus.jinja.sql
```diff
@@ -5,7 +5,7 @@ SELECT csmi.*
         , c.concept_name
         , c.concept_class_id
 FROM {{schema}}cset_members_items csmi
-JOIN concept c ON csmi.concept_id = c.concept_id);
+JOIN {{schema}}concept c ON csmi.concept_id = c.concept_id);
```
joeflack4, line 8: ## Bug fix: `cset_members_item_plus` sometimes dropping
What happened: When the backend tests run, it does `remake_test_schema()`. However, since `{{schema}}` was missing here, when it created this view, it used the schema of the connection, which was set to `test_schema`. So, even though `cset_members_item_plus` was being inserted into `n3c`, it was _dependent on `test_n3c.concept`_. Thus, the next time `remake_test_schema()` ran, during the step at the beginning where it drops w/ cascade `test_n3c`, it resulted in `n3c.cset_members_items_plus` being dropped as well.

test/utils.py
```diff
@@ -71,7 +71,11 @@ def _copy_over_tables(tables: List[str], schema= SCHEMA, test_schema=TEST_SCHEMA
                     END $$;""")
     # Create views
     with get_db_connection(schema=test_schema) as con:
-        refresh_derived_tables_exec(con, list_views(), schema=test_schema)
+        views: List[str] = list_views()
```
joeflack4, line 74: ## Bug fix: `test_n3c.all_csets_view_new" does not exist`
Problem her was that I was making views in `test_n3c` incorrectly. `refresh_derived_tables_exec()` should be called to refresh derived tables, not views. Not sure why this wasn't causing an error before. It may be because `cset_members_items_plus` was the only view and it wasn't really being used.

enclave_wrangler/objects_api.py
```diff
@@ -519,8 +520,8 @@ def fetch_cset_and_member_objects(
         return
 
     # Expression items & concept set members
-    expression_items = []
-    member_items = []
+    expression_items = set()
+    member_items = set()
```
joeflack4, line 524: ## Bug fix: `UniqueViolation: duplicate key value violates unique constraint "concept_pkey1"`
I got the error at first and then couldn't replicate. I think I know what happened.

Previously we set it up to add any concepts to `concept` table. However, there haven't been any new concepts in a long time until the vocab update that just happened.

Now that there are new concepts, it is trying to add to the DB. There _is_ a check where it first sees if the row is in the DB. If it is, it skips adding. That check was working correctly.

So why did this happen?
I think it's because there were _multiple_ instances of the new concept being used. So when it checked, the concept wasn't in the DB. Then, it added the concept once (as it was a member of 1 csets). However, it _kept trying to add_ the same concept again because it was a member of other csets.

How I fixed
I believe this is fixed now by changing it so that when concept members are being identified, it adds them to a set, rather than a list. This should eliminate duplicates, so it will now only try to insert once.

backend/db/ddl-12-all_csets_view.jinja.sql
joeflack4, line 1: ## Bug fix: `cannot drop table all_csets_old because other objects depend on it`

PR #883, opened 2024-09-18
Vocab refresh: Update `networkx` graphs

joeflack4:
addresses #977

## Changes
DB Refresh: relationship_graph fix
- Update: When app starts and also every time home route is hit, it will now update relationship_graph if it is out of date.

General updates
- Codestyle and comment updates
- Moved some code that was causing PyCharm debugger to run tests inside of a non-test file.

Tests
- Bug fix: Updated GH action for tests in order to be able to access relationship_graph.pickle


backend/routes/graph.py
```diff
@@ -375,23 +352,26 @@ def create_rel_graphs(save_to_pickle: bool) -> DiGraph:
     return G # , Gu
 
 
-def load_relationship_graph(save_if_not_exists=True):
+def is_graph_up_to_date(graph_path: str = GRAPH_PATH) -> bool:
```
joeflack4, line 355: ## FYI: `is_graph_up_to_date()`

backend/routes/graph.py
```diff
@@ -375,23 +352,26 @@ def create_rel_graphs(save_to_pickle: bool) -> DiGraph:
...
-            G = pickle.load(pickle_file)
-            # while True:
-            #     try:
-            #         chunk = pickle.load(pickle_file)
-            #         G.add_edges_from(chunk)
-            #     except EOFError:
-            #         break  # End of file reached
+    timer(f'loading {graph_path}')
+    up_to_date = True if not update_if_outdated else is_graph_up_to_date(GRAPH_PATH)
+    if os.path.isfile(graph_path) and up_to_date:
```
joeflack4, line 370: If not true, triggers creation / save of new graph.

backend/routes/graph.py
```diff
@@ -406,20 +386,7 @@ def load_relationship_graph(save_if_not_exists=True):
...
-        # def distance_to_root(G, node):
-        #     n = node
-        #     d = 0
-        #     for p in G.predecessors(node):
-        #         if n in G_ROOTS:
-        #             return d
-        #         d += 1
-        #         n = p
-        #     raise Exception(f"can't find root for {node}")
+        REL_GRAPH = load_relationship_graph()
```
joeflack4, line 392: ## Activates: On server start
Since this file gets imported during server start, this line runs.

joeflack4, line 392: ## TODO: Activate on cron or hitting of home route

Sigfried:
@joeflack4, You got rid of some missing-in-between stuff, or just moved it. Yes -- it's not being used right now. I left the code in place in case we end up with problems where get_all_descendants is making the graph too big. It hasn't happened yet. Maybe we should just get rid of all missing-in-between code and leave a comment pointing to commit where it was last present in case we need to restore it someday.

In looking at it, I noticed that filled_gaps doesn't seem to appear in frontend code at all anymore. It does still appear in some test data.

Sigfried:
I'm about to clean this all up.

PR #889, opened 2024-09-25
Bug fix: GitHub action deprecated module

joeflack4:
## Changes
Bug fix: GitHub action changes
- Update: GitHub deprecated certain actions that were being used by our actions: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/. Even though they were just deprecated, it caused an error from our actions still being on the old version.

PR #898, opened 2024-09-26
Bugfix: Dev

joeflack4:


.github/workflows/backend_dev.yml
```diff
@@ -65,6 +65,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: python-app
+          include-hidden-files: true
```
joeflack4, line 68: ## The fix
This is a new param I have to add as of v4.

The original cause:
- https://github.com/jhu-bids/TermHub/pull/889
- Migration from v2 to v4 https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Issue #547, opened 2023-09-07
Grouping: Show/filter certain concepts (`showThoughCollapsed`, `hideThoughExpanded`)


Sigfried:
# How to handle show/filter post-refactor (2024-10-24)

- [x] Just don't do showThoughCollapsed for now
  - #918
- [ ] #921

## ~Old overview~ Going to start redefining this above
~Already allowing filtering on front end of RxNorm Extension concepts and zero-patient-count concepts and allowing collapse of a concept's descendants. @hlehmann17 suggests doing the filtering (at least for RxNorm Extension) server side because download of large concept sets causes timeout and caching issues (#514.)~

This was an optimization issue. The optimization problems are *mostly* solved (try Expand all with Malignant Cancer and it will still crash the browser.) Filtering out RxNorm Extension no longer works on front or backend. Have added Stats and options dialog. Really hard to get it working quite right.

## Hide though expanded (frontend)
Moved this to a new issue:
- [ ] #922

## Show though collapsed (frontend)
- [x] Expand all (does what it should do, expands all but then allows collapsing specific rows; and toggles on click to Collapse all)
- [x] Definition concepts
- [x] Added
- [x] Removed

## Problems
- [ ] Problem: hide occurs later in code so takes precedence
  - Fix so each options change leaves everything else in place; so, if you do Show added added after Hide zero records, it will also show the added concepts with zero records -- Or, should any Hide reset all the Shows and vice versa instead?
- [x] #913

## UI / UX
- [x] #910
- [x] #911
- [ ] #912

### Exclude from fetch (backend, with frontend option somehow)
-  [x] #616. Exclude RxNorm Extension from fetch unless user requests them explicitly

## Related
- #707
~Direct grouping issue:~
- ~#514~

joeflack4:
This may be a duplicate of #514, though this issue is more concise. Solution (3) in that issue has to do w/ filtering of RxNorm extension, etc.

joeflack4:
Not an entire duplicate now, it looks like. What do you think of adding Harold's proposal _"4. General hiding of concepts / subtrees"_ in #514 to this list? E.g. if a cset has over ~2k members, start to hide nodes.

joeflack4:
May be addressed by:
- #909

joeflack4:
Closing this issue in favor of new issues opened (see links in original post).

joeflack4:
Actually re-opening because this is a grouping issue.

PR #909, opened 2024-10-09
Refactoring `displayedRows` stuff, WIP

Sigfried:
## Notes
- Stop saving everything as persistent object props -- especially displayedRows. Makes code harder to understand. Go back to regular functions that return results instead of saving results to object props.
- Shouldn't have to keep arranging rows and repeatedly processing graph option config
- Documented the options and stuff about hiding/showing -- it's complicated and there are a bunch of cases I'm not sure how to handle. But it's a lot easier to understand with the documentation (inline in GraphState:GraphContainer.getDisplayedRows)
- Trying new strategy of laying out all rows first before trying to hide. Not sure if it's a good idea.

## Related
May address:
- #547


Sigfried:
This has Joe's blessing.

Issue #947, opened 2024-11-04
Fail: Frontend tests broken


joeflack4:
## Overview
After merge of #934, [frontend tests](https://github.com/jhu-bids/TermHub/actions/runs/11668823309/job/32489357838) now broken.

## Errors
`TypeError: Cannot read properties of undefined (reading 'forEach')`

Issue #948, opened 2024-11-04
Fail: Backend tests broken


joeflack4:
## Overview
After merge of #934, [backend tests](https://github.com/jhu-bids/TermHub/actions/runs/11668823313/job/32489358694) now broken.

I'm not sure why they started failing once this was merged, because when I look at the PR page, it seems to show that it was passing.

## Errors
```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedColumn) column "item_flags" does not exist
LINE 13:                         CASE WHEN LENGTH(item_flags) > 0 THE...
```

PR #949, opened 2024-11-05
Granular caching

Sigfried:
I'm confused what's going on here. I'm trying to deal with a couple test failures (#947, #948), and thought the tests would be rerun when I committed code to granular-caching, but that's not happening, so creating a pull request. (Which I will probably merge immediately because there's some broken code in develop right now that this should also fix.


Sigfried:
Tests passing again. Fixed #947 and #948

PR #970, opened 2024-11-10
Bug fix: DB refresh table order

joeflack4:
Bug fix: DB refresh table order
- Fixed bug where order of derived tables refreshed was non-deterministic and wrong.

### DIFF
--- a/backend/db/utils.py
+++ b/backend/db/utils.py
@@ -82,52 +82,59 @@ def get_dependent_tables_queue(independent_tables: Union[List[str], str], _filte
     :return: A list in the correct order such that for every entry in the list, any tables that depend on that entry
     will appear further down in the list.
 
-    todo: Replace heuristic w/ a correct algorithm.
+    TODO: Replace heuristic w/ a correct algorithm.
      I originally had no steps 2&3, and only 1&4 combined. But the result was out of order. This algorithm below is
      based on a quick (but messy/long) heuristic. Basically, the longer dependency trees go first. This corrected the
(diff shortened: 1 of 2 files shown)


backend/db/utils.py
```diff
@@ -170,7 +177,8 @@ def refresh_derived_tables_exec(
     t0 = datetime.now()
     for module in ddl_modules_queue:
         t0_2 = datetime.now()
-        print(f' - creating new table/view: {module}...')
+        table_or_view = 'view' if module in views else 'table'
+        print(f' - creating new {table_or_view}: {module}...')
```
joeflack4, line 181: ## Improve: Refresh logs
Now says 'table' if it's refreshing a table or 'view' if view instead of 'table/view'

backend/db/utils.py
```diff
@@ -82,52 +82,59 @@ def get_dependent_tables_queue(independent_tables: Union[List[str], str], _filte
...
-            table_queues2.append(q)
-
-    # 4. Flatten to single list queue of dependent tables
-    for q in table_queues2:
-        for i in q:
-            if i not in final_queue:
-                final_queue.append(i)
+
+    # TODO: Need correct algo. see docstring -> 2024/11/09
+    if independent_tables == CORE_CSET_TABLES:
```
joeflack4, line 97: ## Bug fix (stop gap)
This is a total stop gap because I just don't have time to make this algorithm deterministic. 

It was sometimes returning the wrong order in which to refresh tables.

We actually don't have that many different situations in which we need to order tables currently, so I will just tell it the order in which to refresh the tables.

PR #976, opened 2024-11-18
Bug fix: DB refresh table order - redux

joeflack4:
## Changes
- Fixed the order to the order in which DDL appear.

Follow up for what was a temporary fix that didn't handle all situations:
- #970


backend/db/utils.py
```diff
@@ -871,6 +872,36 @@ def list_tables(con: Connection = None, schema: str = None, filter_temp_refresh_
...
+def get_ddl_paths_by_name(module_names: Union[List[str], str]) -> List[str]:
+    """Given a list of module/table names, get the paths to each of the DDL files."""
+    modules: List[str] = [module_names] if isinstance(module_names, str) else module_names
+    paths: List[str] = glob(DDL_JINJA_PATH_PATTERN)
+    if modules:
+        paths = [p for p in paths if any([m == os.path.basename(p).split('-')[2].split('.')[0] for m in modules])]
+    return sorted(paths, key=lambda x: int(os.path.basename(x).split('-')[1]))
+
+
+def order_modules_by_ddl_order(module_names: Union[List[str], str]) -> List[str]:
```
joeflack4, line 884: Handled by this new function.

Was able to refactor and re-use some existing code.

Issue #874, opened 2024-09-07
New token for UNITEConceptSetBulk@nih.gov


joeflack4:
## Overview
We got an email that the token for UNITEConceptSetBulk@nih.gov appeared in a public repository and was thus deleted for security purposes. We need a new token.

## Sub-tasks
- [x] 1. Get new token
- [x] 2. Update envs (follow instructions [here](https://github.com/jhu-bids/TermHub/blob/develop/docs/developer.md#updating-environmental-variables), [here](https://github.com/jhu-bids/TermHub/blob/develop/docs/developer.md#updating-environmental-variables), and [here](https://github.com/jhu-bids/TermHub/blob/develop/docs/developer.md#updating-environmental-variables))
  - I'm not sure which environmental variable the "UNITEConceptSetBulk@nih.gov token" represents. If it is `PALANTIR_ENCLAVE_AUTHENTICATION_BEARER_TOKEN`, then I think we need to update everything. Otherwise, not sure.
 - [x] 3. Find out how it was leaked and prevent. (addressed by #980)
  - Search TermHub for plain text keys and values for secret(s) in `.env` file.
    - Make sure it isn't appearing in plain text as the result of any verbose print statements. If so, it's possible that it could appear in GitHub action logs.
  - If it wasn't TermHub, bring it up to BIDS and/or DLs at meeting & email.

joeflack4:
Results of asking codebase LLM (Cursor running Claude) to check:

<details><summary>Details</summary>
<p>

I've analyzed the codebase for potential security risks related to the `PALANTIR_ENCLAVE_AUTHENTICATION_BEARER_TOKEN`. Here are my findings:

1. Direct references to the token:
The token is referenced in several files, but it's not directly exposed in plain text. It's typically accessed through environment variables or configuration files.

2. Printing or logging:
There are no instances where the token is directly printed or logged. However, there are some areas that might indirectly expose the token:


```125:142:enclave_wrangler/utils.py
def log_debug_info():
    """Logs additional info when making HTTP requests"""
    # These two lines enable debugging at httplib level (requests->urllib3->http.client)
    # You will see the REQUEST, including HEADERS and DATA, and RESPONSE with HEADERS but without DATA.
    # The only thing missing will be the response.body which is not logged.
    try:
        import http.client as http_client
    except ImportError:
        # Python 2
        import httplib2 as http_client
    http_client.HTTPConnection.debuglevel = 1

[...]

PR #980, opened 2024-11-18
Security: Stop printing auth token

joeflack4:
addresses: #874


enclave_wrangler/utils.py
```diff
@@ -71,9 +73,9 @@ def get_headers(personal=False, content_type="application/json", for_curl=False,
         headers["Content-type"] = "application/json"
 
     # set_auth_token_key(current_key)
+    # Security: space added after '$ ' to prevent possible accidental interpolation
     if for_curl:
-        # headers["authorization"] = '$' + TOKEN_KEY
-        headers["authorization"] = auth_token
+        headers["authorization"] = '$ ' + TOKEN_KEY
```
joeflack4, line 78: ## Only known offending case
@Sigfried Can you think of any other cases where this is getting printed?

@Sigfried What do you think about this update? It still returns a curl, but you manually have to edit the `' '` out of `'$ '` for added security.

PR #987, opened 2024-11-25
Sync Draft Expression Items

joeflack4:
## Changes
Sync Draft Expression Items
- Now syncing added or deleted expression items while csets are drafts or are finalized.

General
- Bug fix: When adding expression items, values for 'annotation' and 'sourceApplciation' fields were being dropped.

### DIFF
--- a/enclave_wrangler/models.py
+++ b/enclave_wrangler/models.py
@@ -281,7 +281,9 @@ add_mappings(
        includeMapped,                includeMapped
        isExcluded,                   isExcluded
        createdBy,                    created_by
-       createdAt,                    created_at""")
+       createdAt,                    created_at
+       annotation,                   annotation
+       sourceApplication,            source_application""")
 
 # OMOPConceptSet (Version): object <-> dataset
 add_mappings(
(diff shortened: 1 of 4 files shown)


enclave_wrangler/models.py
```diff
@@ -281,7 +281,9 @@ def add_mappings(csv_str: str):
        includeMapped,                includeMapped
        isExcluded,                   isExcluded
        createdBy,                    created_by
-       createdAt,                    created_at""")
+       createdAt,                    created_at
+       annotation,                   annotation
+       sourceApplication,            source_application""")
```
joeflack4, line 286: ## Bug fix: missing `annotation` and `sourceApplication`
These weren't getting imported for `concept_set_version_item` because they were missing from this mapping table.

enclave_wrangler/objects_api.py
```diff
@@ -1123,6 +1122,60 @@ def update_cset_metadata_from_objs(
         conn.close()
 
 
+def sync_cset_expression_changes(cset: Dict[str, Any], con: Connection = None):
```
joeflack4, line 1125: ## Sync cset expression additions/deletions
This is now being done.

## Updates
I wanted to do updates as well, but there was no time for that.

This means that when we refresh members items, they'll be correct, but the expression definitions might not be if they change between initial fetch and finalization.

Issue #999, opened 2024-11-27
Synchronize expression items for problematic csets


joeflack4:
## Overview
In #998, we fixed an issue where the expression items for a cset were incorrect in that some csets were added that were not in the DB, and some were deleted that were still in the DB. For future csets, this will not be a problem as it is addressed by #987.

However, it's very likely that there are other csets in the DB which have this issue.

So, what's required is a 1 time fix for all affected csets.

## Options
a. identify all cases where this issue exists via some query and fix only those csets
b. just sync all csets created after we started importing drafts

I'm going to do this, since it requires very minimal work for me to do, and shouldn't create any problems.

## Additional info
**What this PR will not address:**
Unfortunately, while #987 handled expression item additions/deletions, it did not handle updates. So, if `includeDescendants`, etc, values changed for any expression item after a cset was initially fetched as a draft and before it was finalized, these changes will not be reflected in TermHub. That will be handled by:
- #989

joeflack4:
Done a few days ago.

PR #1001, opened 2024-11-27
Synchronize expression items for problematic csets

joeflack4:
addresses #999

## Changes
- Addresses issue where expressions could be out of sync between the enclave and DB, i.e. some may have been added or deleted since when they became drafts and when they were imported into the DB, and the DB did not get those udpates.


backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -54,6 +54,20 @@ def _ad_hoc_detect_and_delete_discarded_drafts(schema: str = SCHEMA, local: bool
         handle_discarded_drafts(con, csets_and_members, set(draft_ids), report_success=False, use_local_db=local)
 
 
+def _ad_hoc_sync_draft_expression_items(since='2023-08-15', schema=SCHEMA, local=False):
```
joeflack4, line 57: Completed successfully, syncing all expression items since this date, which is when we added the drafts feature.

PR #1007, opened 2024-12-04
GH Action updates: Vocab refresh & Backend tests

joeflack4:
## Changes
Vocab refresh
- Now errors when completes successfully so that it will notify that a restart of the server is necessary to refresh networkx graph.

Backend tests action
- Set cronjob to 1x/day, as this will test database integrity, including for vocab refreshes, which is helpful since due to the way I'm having it error and notify, it isn't running the test.

### DIFF
--- a/.github/workflows/test_backend_e2e_and_unit_and_qc.yml
+++ b/.github/workflows/test_backend_e2e_and_unit_and_qc.yml
@@ -10,6 +10,8 @@ on:
     branches: [ develop, main ]
   pull_request:
     branches: [ develop, main ]
+  schedule:
+    - cron: '0 14 * * *'  # every day 2pm GMT (9/10am EST/EDT)
 
 jobs:
   test:
--- a/backend/db/refresh_dataset_group_tables.py
+++ b/backend/db/refresh_dataset_group_tables.py
@@ -31,6 +31,11 @@ from enclave_wrangler.config import DATASET_GROUPS_CONFIG
 from enclave_wrangler.datasets import download_datasets, get_datetime_dataset_last_updated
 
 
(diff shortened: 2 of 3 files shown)


.github/workflows/test_backend_e2e_and_unit_and_qc.yml
```diff
@@ -10,6 +10,8 @@ on:
     branches: [ develop, main ]
   pull_request:
     branches: [ develop, main ]
+  schedule:
+    - cron: '0 14 * * *'  # every day 2pm GMT (9/10am EST/EDT)
```
joeflack4, line 14: ## Backend tests action: cron
- Set cronjob to 1x/day, as this will test database integrity, including for vocab refreshes, which is helpful since due to the way I'm having it error and notify, it isn't running the test.

backend/db/refresh_dataset_group_tables.py
```diff
@@ -109,6 +114,14 @@ def refresh_dataset_group_tables(
             # DB Counts
             print('Updating database counts. This could take a while...')
             counts_update(f'DB refresh: {",".join(dataset_group)}', schema)
+
+        # Vocab refresh only
+        # todo: ideally this notification would happen differently. More comments near bottom of refresh_voc.yml
+        if group_name == 'vocab' and len(dataset_group) == 1:  # Will be the case for GH actions
+            raise NotAnError('Notification: Restart deployment backend.\nThis is not an error. It is only being raised '
```
joeflack4, line 121: ## Vocab refresh: notification on success
- Now errors when completes successfully so that it will notify that a restart of the server is necessary to refresh networkx graph.

PR #1023, opened 2025-07-24
Trying to fix 0 member refresh with claudes help

Sigfried:


enclave_wrangler/objects_api.py
```diff
@@ -542,7 +542,10 @@ def fetch_cset_and_member_objects(
                     call_github_action('resolve-fetch-failures-excess-items')
         # - fetch member items
         try:
-            cset['member_items']: List[Dict] = get_concept_set_version_members(version_id, return_detail='full')
+            # let's not bother getting members if it's a draft
+            cset['member_items']: List[Dict] = []
+            if not cset['properties']['isDraft']:
+                cset['member_items']: List[Dict] = get_concept_set_version_members(version_id, return_detail='full')
```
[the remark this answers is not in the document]
joeflack4: Hmm, perhaps it would help to do the following in the for loop above the try/except blocks:

```py
cset['member_items']: List[Dict] = []
 cset['expression_items']: List[Dict] = []
```

.github/workflows/resolve_fetch_failures_0_members.yml
```diff
@@ -1,15 +1,19 @@
 name: "Resolve fetch failures: New csets w/ 0 members"
 
 on:
-  schedule:
-    - cron: '10 5 * * *'  # every day 5:10am GMT (12:10/1:10am EST/EDT)
+  # AI-generated: Removed the daily schedule since this is already triggered by refresh.py every 20 minutes
+  # schedule:
+  #   - cron: '10 5 * * *'  # every day 5:10am GMT (12:10/1:10am EST/EDT)
```
joeflack4, line 6:
```suggestion
     - cron: '10 5 * * *'  # every day 5:10am GMT (12:10/1:10am EST/EDT)
```

I would re-activate this 20 minute action.

Rationale:
Per our discussion, the normal DB refresh only kicks off this draft finalization script if it encounters a new draft. This is hypothetical, but let's say that for some reason it's days or weeks before someone makes any new drafts, but in that time frame, existing drafts were finalized. If you deactivate the `cron` on this action, then TermHub will not know about that.

Indeed, imagine if you or one of our colleagues made a draft, but didn't have time to finalize it until a few days later. After you eventually finalize it, who knows when you'll actually get to see it in TermHub. Unless you execute the action manually every time. But do you really want to have to do that? Especially considering other of our colleagues who will not remember or don't have permission to execute our actions.

Keeping this action running regularly ensures that finalized drafts consistently show up ASAP in TermHub.

.github/workflows/resolve_fetch_failures_0_members.yml
```diff
@@ -1,15 +1,19 @@
...
   workflow_dispatch:
     inputs:
       # todo: support list of IDs (comma-delimited)
       version_id:
         description: 'Optional: Concept set version ID'
         required: false
 #        type: string  # integer not supported; type will be corrected by script
+  # AI-generated: Allow triggering from other workflows (like refresh.py does)
+  repository_dispatch:
```
joeflack4, line 15: Hmm, I'm surprised this wasn't there before. I feel like given that this was missing, there should have been an errors in the past when the refresh tried to kick off this script.

.github/workflows/resolve_fetch_failures_0_members.yml
```diff
@@ -1,15 +1,19 @@
...
+  # schedule:
+  #   - cron: '10 5 * * *'  # every day 5:10am GMT (12:10/1:10am EST/EDT)
   workflow_dispatch:
     inputs:
       # todo: support list of IDs (comma-delimited)
       version_id:
         description: 'Optional: Concept set version ID'
         required: false
 #        type: string  # integer not supported; type will be corrected by script
+  # AI-generated: Allow triggering from other workflows (like refresh.py does)
```
joeflack4, line 14:
```suggestion
  # Allow triggering from other workflows (like refresh.py does)
```

Maybe not a bad idea temporarily, but very soon (if not already), so much is going to be written by AI that it's probably moot to have this clarification.

backend/db/resolve_fetch_failures_0_members.py
```diff
@@ -168,17 +175,33 @@ def filter_cset_id_where_0_expanded_members(cset_ids: List[int], schema=SCHEMA,
...
+        # AI-generated: Get csets with 0 expressions
+        zero_expr_rows: List[List[float]] = sql_query(con, qry_zero_expressions, return_with_keys=False)
+        zero_expr_ids: List[int] = [int(x[0]) for x in zero_expr_rows]
+        
+        # AI-generated: Get csets with all excluded expressions
+        all_excluded_rows: List[List[float]] = sql_query(con, qry_all_excluded, return_with_keys=False)
+        all_excluded_ids: List[int] = [int(x[0]) for x in all_excluded_rows]
+        
+        # AI-generated: Return union of both sets
+        return list(set(zero_expr_ids + all_excluded_ids))
```
joeflack4, line 204: Ah, good good. That should fix the 0 expressions issue.

enclave_wrangler/objects_api.py
```diff
@@ -542,7 +544,9 @@ def fetch_cset_and_member_objects(
                     call_github_action('resolve-fetch-failures-excess-items')
         # - fetch member items
         try:
-            cset['member_items']: List[Dict] = get_concept_set_version_members(version_id, return_detail='full')
+            # let's not bother getting members if it's a draft
```
joeflack4, line 547: That makes a lot of sense:
>             # let's not bother getting members if it's a draft

joeflack4 (approved):
The only thing I disapprove of is deactivating the 20 minute thing, but it is your choice!

FYI I can review these future ones, but if you could add to the OP the list of issues this is meaning to address (or otherwise a bulleted list of changes/fixes if no issues exist) that would make it much easier for me!

.github/workflows/resolve_fetch_failures_0_members.yml
Sigfried: Is that cron entry for every 20 minutes or every day at 5:10am?

Sigfried: Also, it looks to me like refresh.py is going to launch resolve_fetch_failures_0_members every 20 minutes no matter what. Am I wrong?

image

